PR #1, opened 2023-11-01
Add new method

TheMysteriousStranger90:
Add new method which will allow to set or change a custom alphabet after instantiating the class. This can be useful for dynamically adjusting the encoding.

### DIFF
--- a/src/.idea/.idea.Microsoft.Security.Utilities/.idea/.gitignore
+++ b/src/.idea/.idea.Microsoft.Security.Utilities/.idea/.gitignore
@@ -1,0 +1,13 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Rider ignored files
+/modules.xml
+/contentModel.xml
+/projectSettingsUpdater.xml
+/.idea.Microsoft.Security.Utilities.iml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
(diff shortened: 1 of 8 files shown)


TheMysteriousStranger90:
@microsoft-github-policy-service agree

src/Microsoft.Security.Utilities/CustomAlphabetEncoder.cs
```diff
@@ -115,5 +115,41 @@ public byte[] Decode(string encodedValue)
...
+        
+        /// <summary>
+        /// Sets a custom alphabet to be used for all encoding and decoding operations.
+        /// </summary>
+        /// <param name="customAlphabet">A string representing the custom alphabet. It must consist of non-whitespace ASCII characters, including letters, numbers, and/or punctuation.</param>
+        /// <exception cref="ArgumentException">Throws an exception if the alphabet contains duplicates or contains forbidden characters.</exception>
+        /// <remarks>
+        /// This method allows dynamic changes to the alphabet used for encoding and decoding data. Before setting a new alphabet, the method performs checks for duplicates and the presence of forbidden characters.
+        /// </remarks>
+        public void SetCustomAlphabet(string customAlphabet)
```
michaelcfanning, line 127: >SetCustomAlphabet [](http://example.com/codeflow?start=20&length=17)

Sorry for not noticing this PR earlier! This looks like a good idea. It'd be best to share the argument validated logic between this method and the constructor via a shared helper.

src/.idea/.idea.Microsoft.Security.Utilities/.idea/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Default ignored files
```
michaelcfanning, line 1: >ignored [](http://example.com/codeflow?start=10&length=7)

maybe add `.idea/` as well? you are an intellij user?

PR #8, opened 2024-05-09
Add `IdentifiableSecrets.ComputeChecksumSeed`

michaelcfanning:
- NEW: Add `IdentifiableSecrets._ComputeChecksumSeed` to derive checksum seeds from versioned string literals, e.g., `ReadKey0`.

These literals must be 8 characters with at least one trailing digit.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -78,6 +79,33 @@ public static bool IsBase64UrlEncodingChar(this char ch)
...
+    public static ulong ComputeChecksumSeed(string versionedKeyKind, out string hexFormatted)
+    {
+        hexFormatted = null;
+
+        if (versionedKeyKind == null)
+        {
+            throw new ArgumentNullException(nameof(versionedKeyKind));
+        }
+
+        if (versionedKeyKind.Length != 8 ||
```
rwoll: How is a developer using the API supposed to know there's this implicit requirement/convention of ending with a digit until they hit this error? I suggest adopting Jimmy's API suggestion of making version an explicit argument.

rwoll (dismissed):
Approved, but consider using a more explicit API.

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
michaelcfanning: OK, I misunderstood. So, what's the convention? A string of 7 chars + a single digit that allow for 10 discrete versions?  A string of 6 characters and two digits that allows for 100? Should the API simply create your trailing version and as long as the length of it and the string literal does not exceed 8 chars we are good? It's all looking a little complex. :)

And so I propose to just have people call the API. Let me go back and add some XML comments to clarify the approach.

Also, obviously, feel free to push back or comment on what I've said above.

rwoll: Until this morning, I thought the only requirement for the seed generation was: 8 characters of ASCII text with the suggestion that the text should be human-friendly alluding to the key kind.

Then there was mention of reserving the last character for "version".

Is the version your idea or proposed by someone else? What do you envision "version" being used for?

jimmyca15: I would suggest 

`public static ulong ComputeChecksumSeed(string keyKind, int version)`

Where keyKind between 1 and 6 characters and version is [0,99].
The implementation would add padding to keyKind as necessary.

michaelcfanning: OK! so I can do this. but if users provide a version >99 then this API will throw. :) is that ok? because we think no one is likely to hit this threshold?

Sounds like a reasonable compromise to what I proposed.

jimmyca15: > because we think no one is likely to hit this threshold

That and the original suggestion was `RWSeed00` which allowed for two digit growth. I have no opinion on how much version growth should be allowed.

PR #14, opened 2024-05-15
New checks

michaelcfanning:


src/Microsoft.Security.Utilities.Core/HighConfidenceMicrosoftSecurityModels/IdentifiableKey.cs
```diff
@@ -32,7 +34,8 @@ public IdentifiableKey()
             {
                 if (IdentifiableSecrets.ValidateBase64Key(match,
                                                           checksumSeed,
-                                                          Signature))
+                                                          Signature,
+                                                          EncodeForUrl))
```
michaelcfanning, line 38: >EncodeForUrl [](http://example.com/codeflow?start=58&length=12)

Unit test gap TBD

PR #15, opened 2024-05-15
Add new scan types and fix UTF16 trailing byte case

beaubelgrave:
Add new 32, 39, and 40 byte signatures.
Add AAD signatures (7Q~ and 8Q~).
Fix UTF16 trailing byte case.

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
@@ -486,8 +486,8 @@ impl PossibleScanMatch {
                     ScanMatch::new(
                         ScanMatchType::His32Utf16,
                         start,
-                        (count * 2) as u64,
-                        &bytes[0..count],
+                        (len * 2) as u64,
+                        &bytes[0..len],
                         want_text))
             },
 
@@ -511,8 +511,8 @@ impl PossibleScanMatch {
                     ScanMatch::new(
                         ScanMatchType::His64Utf16,
                         start,
-                        (count * 2) as u64,
-                        &bytes[0..count],
+                        (len * 2) as u64,
+                        &bytes[0..len],
                         want_text))
             },
 
@@ -552,8 +552,8 @@ impl PossibleScanMatch {
                     ScanMatch::new(
                         ScanMatchType::His39Utf16,
                         start,
-                        (count * 2) as u64,
-                        &bytes[0..count],
+                        (len * 2) as u64,
+                        &bytes[0..len],
                         want_text))
             },
 
@@ -593,8 +593,8 @@ impl PossibleScanMatch {
                     ScanMatch::new(
                         ScanMatchType::His40Utf16,
                         start,
-                        (count * 2) as u64,
-                        &bytes[0..count],
+                        (len * 2) as u64,
+                        &bytes[0..len],
                         want_text))
             },
 
@@ -634,8 +634,8 @@ impl PossibleScanMatch {
                     ScanMatch::new(
                         ScanMatchType::HisA7Utf16,
                         start,
-                        (count * 2) as u64,
-                        &bytes[0..count],
+                        (len * 2) as u64,
+                        &bytes[0..len],
                         want_text))
             },
 
@@ -675,8 +675,8 @@ impl PossibleScanMatch {
                     ScanMatch::new(
                         ScanMatchType::HisA8Utf16,
                         start,
-                        (count * 2) as u64,
-                        &bytes[0..count],
+                        (len * 2) as u64,
+                        &bytes[0..len],
                         want_text))
             },
 
@@ -1379,6 +1379,9 @@ mod tests {
         cases.push(Case::new(
             "zzz8Q~zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz",
             "zzz8Q~zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz"));
+        cases.push(Case::new(
+            "zzz8Q~zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzZZZ",
+            "zzz8Q~zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz"));
 
         /* 32-byte cases */
         cases.push(Case::new(
@@ -1408,6 +1411,12 @@ mod tests {
         cases.push(Case::new(
             "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbAZEGLiQbng=",
             "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbAZEGLiQbng="));
+        cases.push(Case::new(
+            "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbAZEGLiQbng=ZZZZ",
+            "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbAZEGLiQbng="));
+        cases.push(Case::new(
+            "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbAZEGLiQbngZZZZ",
+            "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbAZEGLiQbng"));
 
         /* 39-byte cases */
         cases.push(Case::new(
@@ -1419,6 +1428,9 @@ mod tests {
         cases.push(Case::new(
             "dddddddddddddddddddddddddddddddddddddddddd+ACRCUDxQE",
             "dddddddddddddddddddddddddddddddddddddddddd+ACRCUDxQE"));
+        cases.push(Case::new(
+            "dddddddddddddddddddddddddddddddddddddddddd+ACRCUDxQEZZZZZ",
+            "dddddddddddddddddddddddddddddddddddddddddd+ACRCUDxQE"));
 
         /* 40-byte cases */
         cases.push(Case::new(
@@ -1427,6 +1439,9 @@ mod tests {
         cases.push(Case::new(
             "ddddddddddddddddddddddddddddddddddddddddddddAzFu182vhA==",
             "ddddddddddddddddddddddddddddddddddddddddddddAzFu182vhA=="));
+        cases.push(Case::new(
+            "ddddddddddddddddddddddddddddddddddddddddddddAzFu182vhAZZZZZ",
+            "ddddddddddddddddddddddddddddddddddddddddddddAzFu182vhA"));
 
         /* 64-byte cases */
         cases.push(Case::new(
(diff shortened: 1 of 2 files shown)


src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +126,148 @@ impl PossibleScanMatch {
...
+            ScanMatchType::HisA7Utf8 => { HIS_A7_UTF8_LEN },
+            ScanMatchType::HisA7Utf16 => { HIS_A7_UTF16_LEN },
+            ScanMatchType::HisA8Utf8 => { HIS_A8_UTF8_LEN },
+            ScanMatchType::HisA8Utf16 => { HIS_A8_UTF16_LEN },
         }
     }
 
     pub fn match_type(&self) -> &ScanMatchType { &self.mtype }
 
+    fn his_a7_matched_bytes(data: &[u8]) -> usize {
```
michaelcfanning, line 154: >his_a7_matched_bytes [](http://example.com/codeflow?start=7&length=20)

I believe this match and the a8 will not fire for secrets that contain the '~' and '.' characters, correct? both are legal for AAD SPN passwords. #Closed

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +126,148 @@ impl PossibleScanMatch {
...
+        }
+
+        for b in &data[6..37] {
+            if !b.is_base64() {
+                return 0;
+            }
+        }
+
+        if data.len() >= HIS_A7_UTF8_LEN {
+            if data[37] == b'=' {
```
michaelcfanning: >if data[37] == b'=' [](http://example.com/codeflow?start=12&length=19)

do we need this? this secret type is not base64-encoded and won't have a '=' terminator.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +126,148 @@ impl PossibleScanMatch {
...
+                return HIS_A7_UTF8_LEN;
+            }
+        }
+
+        37
+    }
+
+    fn his_a8_matched_bytes(data: &[u8]) -> usize {
```
michaelcfanning, line 179: >his_a8_matched_bytes [](http://example.com/codeflow?start=7&length=20)

same notes here as for `a7`

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Good catch, fixed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Removed and pushed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Fixed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+            ScanMatchType::HisA8Utf8 => { HIS_A8_UTF8_LEN },
+            ScanMatchType::HisA8Utf16 => { HIS_A8_UTF16_LEN },
         }
     }
 
     pub fn match_type(&self) -> &ScanMatchType { &self.mtype }
 
+    fn his_a7_matched_bytes(data: &[u8]) -> usize {
+        /*
+         * 3 Base64 + 3 signature + 31 Base64 + optional 1 [=]
```
michaelcfanning: > + optional 1 [=] [](http://example.com/codeflow?start=45&length=17)

nit: this comment is no longer correct

michaelcfanning: 3 url unreserved + 3 signature + 31 url unreserved

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+                return 0;
+            }
+        }
+
+        HIS_A7_UTF8_LEN
+    }
+
+    fn his_a8_matched_bytes(data: &[u8]) -> usize {
+        /*
+         * 3 Base64 + 3 signature + 34 Base64 + optional 1 [=]
```
michaelcfanning: >3 Base64 + 3 signature + 34 Base64 + optional 1 [=] [](http://example.com/codeflow?start=11&length=51)

comment s/b:
3 url unreserved + 3 signature + 34 url unreserved

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+                return 0;
+            }
+        }
+
+        HIS_A8_UTF8_LEN
+    }
+
+    fn his_39_matched_bytes(data: &[u8]) -> usize {
+        /*
+         * 42 Base64 + 4 signature + 1 [A-D] + 5 Base64 + optional 1 [=]
```
michaelcfanning: >+ optional 1 [=] [](http://example.com/codeflow?start=56&length=16)

a 39 byte secret won't encode a trailing '=' sign. so i'm not sure we require that processing here.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+        }
+
+        for b in &data[47..52] {
+            if !b.is_base64() {
+                return 0;
+            }
+        }
+
+        if data.len() >= HIS_39_UTF8_LEN {
+            if data[52] == b'=' {
```
michaelcfanning: >if data[52] == b'=' [](http://example.com/codeflow?start=12&length=19)

i'm not sure we need these. I'd simply return the match. the follow-on processing can examine the checksum to make sure we've got a correct thing.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -8,16 +8,28 @@ const HIS2_UTF16_LEN: usize = HIS2_UTF8_LEN * 2;
 const HIS2_UTF16_SHORT_LEN: usize = HIS2_UTF8_SHORT_LEN * 2;
 const HIS2_UTF16_SHORT_LEN_BE: usize = HIS2_UTF16_SHORT_LEN - 1;
 
+const HIS_A7_UTF8_LEN: usize = 37;
+const HIS_A8_UTF8_LEN: usize = 40;
 const HIS_32_UTF8_LEN: usize = 44;
+const HIS_39_UTF8_LEN: usize = 53;
```
michaelcfanning: >53 [](http://example.com/codeflow?start=31&length=2)

this should be 52

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+            ScanMatchType::HisA7Utf8 => { HIS_A7_UTF8_LEN },
+            ScanMatchType::HisA7Utf16 => { HIS_A7_UTF16_LEN },
+            ScanMatchType::HisA8Utf8 => { HIS_A8_UTF8_LEN },
+            ScanMatchType::HisA8Utf16 => { HIS_A8_UTF16_LEN },
         }
     }
 
     pub fn match_type(&self) -> &ScanMatchType { &self.mtype }
 
+    fn his_a7_matched_bytes(data: &[u8]) -> usize {
```
michaelcfanning, line 154: >his_a7_matched_bytes [](http://example.com/codeflow?start=7&length=20)

maybe not super important but this specific secret kind does *not* have a checksum. and so this is a case in theory where looking at any leading or trailing characters could help eliminate a false positive. 

it is helpful that the non-base64-encoding `~` character is part of the signature, as the clearest source of noise would be a signature embedded in a lengthier base64-encoded thing. for now, I think we can leave all this as is, just thinking aloud.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+            return 0;
+        }
+
+        for b in &data[0..42] {
+            if !b.is_base64() {
+                return 0;
+            }
+        }
+
+        /* NOTE: We skip the signature since we already checked */
```
michaelcfanning, line 218: >* NOTE: We skip the signature since we already checked */ [](http://example.com/codeflow?start=9&length=57)

NIT: this comment is true in other helpers and it may be useful to add it to others, for code readability.

michaelcfanning: specifically, in `his_a7_matched_bytes` and `his_a8_matched_bytes`

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+        if data.len() >= HIS_39_UTF8_LEN {
+            if data[52] == b'=' {
+                return HIS_39_UTF8_LEN;
+            }
+        }
+
+        52
+    }
+
+    fn his_40_matched_bytes(data: &[u8]) -> usize {
```
michaelcfanning, line 233: >his_40_matched_bytes [](http://example.com/codeflow?start=7&length=20)

Just a note, there are only 3 forms of matching for any HIS v1 encoded thing, which map to key forms require no, two or 4 bits of padding to keep the fixed signature lined up properly in the base64-encoded form. This helper captures one pattern, your 64 byte helper captures another and the 39 byte helper captures the third scenario. we may not require any generalization taking advantage of this fact, since the fact is we don't project any additional HIS v1 keys that aren't 32, 39, 40, or 64 byte in length.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: I thought the correlation hash would use the ='s? If not, then we can remove.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -110,11 +138,136 @@ impl PossibleScanMatch {
...
+        /* NOTE: We skip the signature since we already checked */
+
+        for b in &data[48..53] {
+            if !b.is_base64() {
+                return 0;
+            }
+        }
+
+        match data[53] {
+            b'A' | b'O' | b'g' | b'w' => { },
```
michaelcfanning: >O [](http://example.com/codeflow?start=21&length=1)

I think this character s/be a `Q` not a `O`. This comment suggests I should create a new unit test that spams you with actual generated test keys. If I'm right, your code will fail a small % of time. Let me prove there's a bug, it will be good testing to have. :)

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Will remove just for 39 case

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Will do

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Added

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Added

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Fixed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Good eye! fixed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
michaelcfanning: [a-zA-Z0-9\-_]{22}XXXX[A-D][a-zA-Z0-9\-_]{5}
[a-zA-Z0-9\-_]{24}XXXX[a-zA-Z0-9\-_]{5}[AQgw]
[a-zA-Z0-9\-_]{25}XXXX[A-P][a-zA-Z0-9\\-_]{5}

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
beaubelgrave: Added.

michaelcfanning (approved):
:shipit:

michaelcfanning:
awesome

PR #17, opened 2024-05-16
v1.4.20

michaelcfanning:
# 1.4.20 - 05/16/2024
- BRK: Add `ComputeHash32(byte[], ulong, int, int)` helper to bring .NET framework and .NET core APIs into alignment.
- BRK: Return value of `ISecretMaskerDetectSecrets(string)` is `IEnumerable<Detection>` (not `ICollection`) for best yield iterator compatibility.
- BUG: Honor `url-safe` option in key `GenerateCommand` to produce URL-safe base64-encoded patterns.
- NEW: Update `SEC101_158_AzureFunctionIdentifiableKey1` ,`SEC101_176_AzureContainerRegistryIdentifiableKey`, and `SEC101_190_AzureEventGridIdentifiableKey` to derive from `IdentifiableKey` base.
- NEW: Implement preliminary high-performance `IdentifiableScan` engine that consume Rust library for detections.

This change also drops in some preliminary build scripts. Those build scripts first compile the Rust library and copies them to the `refs` dir (where they will be picked up and copied by all dependent projects).

I also filled out a test that ensures that a large # of generated HIS v1 keys are flagged by the Rust code. We previously had a bug in detection that would only have been reproduced in ~1 in 62 every allocations. 

Also cleaned up a bit of the shared props file expression.


suvamM (approved):

yongyan-gh (approved):
:shipit:

PR #24, opened 2024-05-22
Update `CommonAnnotatedKey` fields scope to be accessed by derived classes

yongyan-gh:
## Description
- Update `CommonAnnotatedKey` private fields `bytes` and `base64Key` to properties with private setter, so their values can be accessed by derived classes.

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -11,6 +11,9 @@
 - FNS => Flase negative reduction in static analysis.
 
 # 1.4.22 - 05/21/2024
+- BUG: Update `CommonAnnotatedKey` private fields `bytes` and `base64Key` to protected fields so they can be accessed by derived classes.
+
+# 1.4.22 - 05/21/2024
 - BUG: Fix `IdentifiableSecrets.ComputeDerivedSymmetricKey` and `IdentifiableSecrets.ComputeDerivedIdentifiableKey` to properly initialize the `HMACSHA256` algorithm with the cask/identifiable secret.
 
 # 1.4.21 - 05/21/2024
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -10,7 +10,7 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
 
-# 1.4.23 - 05/21/2024
+# UNRELEASED
 - BUG: Update `CommonAnnotatedKey` private fields `bytes` and `base64Key` to properties with private setter, so their values can be accessed by derived classes.
 
 # 1.4.22 - 05/21/2024
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -10,7 +10,7 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
 
-# UNRELEASED
+# 1.4.23 - UNRELEASED
 - BUG: Update `CommonAnnotatedKey` private fields `bytes` and `base64Key` to properties with private setter, so their values can be accessed by derived classes.
 
 # 1.4.22 - 05/21/2024
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -10,7 +10,8 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
 
-# 1.4.23 - UNRELEASED
+# UNRELEASED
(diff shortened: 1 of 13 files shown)


docs/ReleaseHistory.md
```diff
@@ -10,6 +10,9 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
 
+# 1.4.22 - 05/21/2024
```
michaelcfanning: >05/21/2024 [](http://example.com/codeflow?start=11&length=10)

this should be UNRELEASED

michaelcfanning: >22 [](http://example.com/codeflow?start=6&length=2)

s/be 1.4.23

src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
```diff
@@ -59,8 +59,10 @@ public static bool TryCreate(string key, out CommonAnnotatedKey secret)
             return true;
         }
 
-        private byte[] bytes;
-        private string base64Key;
+        protected byte[] bytes;
```
michaelcfanning: >protected [](http://example.com/codeflow?start=8&length=9)

protected fields actually aren't legal according to the .NET guidelines!

Instead, create properties, can we just make the setter protected?

public string EncodedKey { get; protected set; }
public byte[] KeyBytes { get; protected set; }

yongyan-gh: got it

yongyan-gh: updated to properties with protected set

docs/ReleaseHistory.md
yongyan-gh: done

yongyan-gh: updated to 1.4.23, correct?

michaelcfanning: Here's your entry. This is going into 1.4.23 but it didn't ship yesterday, it is unreleased. 

`1.4.23 - UNRELEASED`

src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
```diff
@@ -59,33 +59,36 @@ public static bool TryCreate(string key, out CommonAnnotatedKey secret)
             return true;
         }
 
-        private byte[] bytes;
-        private string base64Key;
+        public byte[] KeyBytes { get; protected set; }
```
michaelcfanning: >KeyBytes [](http://example.com/codeflow?start=22&length=8)

The setter for either of these properties should update the other form as well, arguably.

So if you set the bytes, rewrite the encoded key and vice versa. We should probably do that if we expect classes to extend this case class.

version.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.4.22",
+  "version": "1.4.23",
```
michaelcfanning: >23 [](http://example.com/codeflow?start=18&length=2)

btw you don't actually need to publish this package to get this change, just update your submodule for now!

docs/ReleaseHistory.md
yongyan-gh: done

src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
yongyan-gh: updated the setter logic to rewrite another property.

src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
```diff
@@ -59,33 +58,56 @@ public static bool TryCreate(string key, out CommonAnnotatedKey secret)
...
+        }
+
+        protected CommonAnnotatedKey() { }
 
         private CommonAnnotatedKey(byte[] bytes)
         {
-            this.bytes = bytes;
-            this.base64Key = Convert.ToBase64String(this.bytes);
+            this.keyBytes = bytes;
+            this.encodedKey = Convert.ToBase64String(this.keyBytes);
```
michaelcfanning, line 90: >ToBase64String [](http://example.com/codeflow?start=38&length=14)

where are your unit test for this change, friend? :)

yongyan-gh:
close for now, if needed in future will revisit

PR #27, opened 2024-05-23
Scan: Move to generalized scan and named outputs

beaubelgrave:
Move the IdentififableScan pattern in Rust and C# over to the generalized scanning logic.

Each returned match now has a name IE: SEC101/200. Use these names to lookup the existing RegexPattern in C# and redact, etc.


michaelcfanning (approved):
:shipit:

src/Microsoft.Security.Utilities.Core/IdentifiableScan.cs
```diff
@@ -53,59 +34,70 @@ public enum MatchType : ushort
...
         this.generateCorrelatingIds = generateCorrelatingIds;
-        this.signatureToLengthMap = new Dictionary<string, IList<RegexPattern>>();
+        this.idToLengthMap = new Dictionary<string, IList<RegexPattern>>();
+
+        // TODO: This is missing in regexPatterns.
+        this.idToLengthMap["SEC101/200"] = new List<RegexPattern>
+        {
+            new RegexPattern(
+                "SEC101/200",
+                "AzureCommonAnnotatedSecurityKey",
```
michaelcfanning, line 70: >AzureCommonAnnotatedSecurityKey [](http://example.com/codeflow?start=17&length=31)

i will change this name when I add the C# check, coming in a moment.

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
```diff
@@ -27,6 +27,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.Buffers">
+      <Version>4.5.1</Version>
+      <Version>4.5.1</Version>
```
michaelcfanning, line 32: >Version [](http://example.com/codeflow?start=22&length=7)

probably we can delete two of these? :) if so, i'll clean up in a moment.

PR #32, opened 2024-05-27
Benchmark update

michaelcfanning:


src/Microsoft.Security.Utilities.Core/HighConfidenceMicrosoftSecurityModels/IIdentifiableKey.cs
```diff
@@ -11,12 +11,12 @@ public interface IIdentifiableKey
 
         public string Name { get; } 
         
-        public string Signature { get; }
-
         public uint KeyLength { get; }
         
         public bool EncodeForUrl { get; }
 
+        public ISet<string> Signatures { get; }
```
michaelcfanning, line 18: So, we have a general design tension, a single rule can currently incorporate multiple signatures (specifically, AAD SPNs) but a multiple rules may also share a signature (such as the IOT keys). This complicates producing a clean design.

PR #33, opened 2024-05-29
Detection correlating id property

michaelcfanning:
- NEW: Create distinct `Detection.CrossCompanyCorrelatingId` property.

Currently, we have a convention of storing the correlating id along with the regex pattern rule id in the `Detection.RedactionToken` property. This change puts the correlating id itself into its own property. This allows that data to be generated but for users to prefer an alternate redaction token in the masking scenario.

This change also fixes non-determinism in tests, which I'd previously attempted to change but simply used the wrong mechanism to ensure a non-zero/default value for initial property values.


src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -721,7 +730,7 @@ public void SecretMasker_RemoveEncodedSecrets()
     [TestMethod]
     public void SecretMasker_NotAddShortEncodedSecrets()
     {
-        using var secretMasker = new SecretMasker() { MinimumSecretLength = 3 };
+        using var secretMasker = new SecretMasker() { MinimumSecretLength = 3, DefaultRegexRedactionToken = "zzz", DefaultLiteralRedactionToken = "yyy" };
         secretMasker.AddLiteralEncoder(new LiteralEncoder(x => x.Replace("123", "ab")));
```
rwoll, line 739: What's the intended effect of this statement? I see `123` masked by `secretMasker.AddValue("123");`, but no effect of `secretMasker.AddLiteralEncoder(new LiteralEncoder(x => x.Replace("123", "ab")));` in the test. Is that intentional? If so, please add a comment or rename the test indicating what nuance you're trying to test. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -730,10 +739,24 @@ public void SecretMasker_NotAddShortEncodedSecrets()
...
+        Assert.AreEqual("abyyycdyyy", result);
     }
 
 
+    [TestMethod]
+    public void SecretMasker_DistinguishLiteralAndRegexRedactionTokens()
+    {
+        using var secretMasker = new SecretMasker() { MinimumSecretLength = 3, DefaultRegexRedactionToken = "zzz", DefaultLiteralRedactionToken = "yyy" };
+        
+        secretMasker.AddRegex(new RegexPattern("1000", "Name", 0, "abc"));
```
rwoll: can you use named args here? Just had to click through to definition of RegexPattern to understand that the only relevant bit here was the `abc`. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -143,23 +144,33 @@ private void ValidateSecurityModelsMasking(IEnumerable<RegexPattern> patterns, I
...
                             string secretValue = testExample;
                             string moniker = pattern.GetMatchMoniker(secretValue);
-                            string sha256Hash = RegexPattern.GenerateSha256Hash(secretValue);
 
                             // 1. All generated test patterns should be detected by the masker.
                             string redacted = secretMasker.MaskSecrets(secretValue);
                             bool result = redacted.Equals(secretValue);
                             result.Should().BeFalse(because: $"'{secretValue}' for '{moniker}' should be redacted from scan text.");
+
+                            if (testExample.Contains(";"))
```
rwoll, line 157: huh? why is `;` magic? #Resolved

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -143,23 +144,33 @@ private void ValidateSecurityModelsMasking(IEnumerable<RegexPattern> patterns, I
...
+                            {
+                                continue;
+                            }
+
+                            string expectedRedactedValue = generateCrossCompanyCorrelatingIds
+                                ? $"{pattern.Id}:{RegexPattern.GenerateCrossCompanyCorrelatingId(secretValue)}" 
+                                : "+++";
```
rwoll: what does `+++` mean? #Resolved

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -196,11 +196,17 @@ public override int GetHashCode()
...
                     generateCrossCompanyCorrelatingIds && DetectionMetadata.HasFlag(DetectionMetadata.HighEntropy)
-                        ? $"{Id}:{GenerateCrossCompanyCorrelatingId(match.Value)}"
+                        ? GenerateCrossCompanyCorrelatingId(match.Value)
+                        : null;
+
+                string? redactionToken = crossCompanyCorrelatingId != null
+                        ? $"{Id}:{crossCompanyCorrelatingId}"
                         : defaultRedactionToken;
 
+                redactionToken ??= FallbackRedactionToken;
```
rwoll: Should `s/FallbackRedactionToken/defaultRedactionToken`? It looks like not since the ternary already covers falling back to `defaultRedactionToken` and then this line will only trigger if Ccid and defaultRedactionToken is null.

src/Tests.Microsoft.Security.Utilities.Core/DetectionTests.cs
```diff
@@ -18,21 +18,22 @@ public void Detection_ConstructorValuesSet()
...
             string name = $"{Guid.NewGuid()}";
-            string sha256Hash = $"{Guid.NewGuid()}";
+            string crossCompanyCorrelatingId = $"{Guid.NewGuid()}";
             string redactionToken = $"{Guid.NewGuid()}";
 
             var metadata = (DetectionMetadata)0B_11111;
-            int start = Math.Min(1, (int)DateTime.UtcNow.Ticks % 99);
-            int length = Math.Min(1, (int)DateTime.UtcNow.Ticks % 99);
-            TimeSpan rotationPeriod = TimeSpan.FromSeconds(Math.Min(1, DateTime.UtcNow.Second));
+            int start = Math.Max(1, (int)DateTime.UtcNow.Ticks % 99);
```
rwoll, line 25: :)

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
rwoll: I see it now in another file, but maybe add an inline comment? *Not* suggesting replacing this instance with a const here, since it's nice this test will break if someone changes the const making them aware of the "fallout".

rwoll (changes requested):

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
michaelcfanning: So this is ancient stuff that originates with the ADO oss masking code. The intent of what I added was to demonstrate specifically that the literal replacement token `yyy` was preferred and the regex redaction token `zzz` is not in play. It's not really cricket to overload a test like this to get this additional test value. So I can just create a new discrete test.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
michaelcfanning: yes indeed, my usual practice, lapsed here, sorry.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -730,10 +739,24 @@ public void SecretMasker_NotAddShortEncodedSecrets()
         var input = "ab123cd345";
         var result = secretMasker.MaskSecrets(input);
 
-        Assert.AreEqual("ab***cd***", result);
+        Assert.AreEqual("abyyycdyyy", result);
     }
 
 
+    [TestMethod]
+    public void SecretMasker_DistinguishLiteralAndRegexRedactionTokens()
```
michaelcfanning, line 769: >SecretMasker_DistinguishLiteralAndRegexRedactionTokens [](http://example.com/codeflow?start=16&length=54)

oh, look, i *did* author a test ensuring that these two classes of redaction token are honored. :)

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
michaelcfanning: // TODO: the generated examples don't distinguish the secret from surrounding context,
                            // for rules such as our connection string detecting logic. We need a future change to
                            // separate this data. For now, we skip all connection string patterns. This is a problem
                            // for masking only (and not detection) because we have no location details when masking.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
michaelcfanning: Right, what to do here, good question. I'm favoring using the product const string `RegexPattern.FallbackRedactionToken`. The reason is that we have lots of other data-driven tests with encoded versions of this that will be good signal if that const changes. using the product const here will DRY things out a bit.

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
michaelcfanning: right so this is odd. basically, the code is saying, 'you can set defaultRedactionToken to null but if you do, we will use the fallback redaction token'

alternately, i could have put a null check at the beginning of the API forbidding using a null value.

michaelcfanning: Let me get your thoughts on approach, then I'll put in this change next PR. Minimally this should be commented doing that now.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -150,14 +150,18 @@ private void ValidateSecurityModelsMasking(IEnumerable<RegexPattern> patterns, I
...
+                            // for masking only (and not detection) because we have no location details when masking.
                             if (testExample.Contains(";"))
                             {
                                 continue;
                             }
 
                             string expectedRedactedValue = generateCrossCompanyCorrelatingIds
                                 ? $"{pattern.Id}:{RegexPattern.GenerateCrossCompanyCorrelatingId(secretValue)}" 
-                                : "+++";
+                                : RegexPattern.FallbackRedactionToken;
```
rwoll, line 164: I think what you had before was fine, just needed an inline comment along the lines of `+++ is the fallback redaction token`.

By swapping with a reference to a variable this test, unfortunately, will not break if you unintetionally change the redaction token var not realizing all the places it surfaces.

rwoll (dismissed):

PR #34, opened 2024-05-29
chore: avoid packing Benchmark project

rwoll:
There's no need to pack the Benchmark project,
and doing so slows down the build as well produces extra NuGet packages that most be avoid lest they risk being published if wildcard matching is used.


src/Microsoft.Security.Utilities.Benchmarks/Microsoft.Security.Utilities.Benchmarks.csproj
```diff
@@ -6,6 +6,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <LangVersion>latest</LangVersion>
+    <IsPackable>false</IsPackable>
```
michaelcfanning, line 9: >false [](http://example.com/codeflow?start=16&length=5)

the default is true? nuts

michaelcfanning (approved):
:shipit:

PR #35, opened 2024-05-30
Harden `IdentifiableSecrets.TryValidateCommonAnnotatedKey` for invalid inputs.

michaelcfanning:
- BUG: Harden `IdentifiableSecrets.TryValidateCommonAnnotatedKey` for a variety of invalid inputs.

Partners report to us that this `Try` pattern API throws exceptions for invalid inputs. Now fixed.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -55,6 +55,31 @@ public static class IdentifiableSecrets
     public static bool TryValidateCommonAnnotatedKey(string key,
                                                      string base64EncodedSignature)
     {
+        if (string.IsNullOrWhiteSpace(key))
+        {
+            return false;
+        }
+
+        try
+        {
+            ValidateCommonAnnotatedKeySignature(base64EncodedSignature);
+        }
+        catch (ArgumentException)
+        {
+            return false;
+        }
+
+        if (key.Length != 84 && key.Length != 86)
+        {
+            return false;
+        }
+
+        string signature = key.Substring(76, 4);
+
+        string alternate = char.IsUpper(signature[0]) 
+            ? signature.ToLowerInvariant() 
+            : signature.ToUpperInvariant();
+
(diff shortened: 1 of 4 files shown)


src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -29,6 +30,91 @@ static IdentifiableSecretsTests()
 
         private static string s_base62Alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890";
 
+        [TestMethod]
```
rwoll, line 33: Awesome tests! This is a prime target function for a fuzzer.

What happens if you pass keys with emoji, null bytes, etc?

rwoll (approved):

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -55,6 +55,31 @@ public static bool IsBase64UrlEncodingChar(this char ch)
...
+        try
+        {
+            ValidateCommonAnnotatedKeySignature(base64EncodedSignature);
+        }
+        catch (ArgumentException)
+        {
+            return false;
+        }
+
+        if (key.Length != 84 && key.Length != 86)
```
rwoll, line 72: nit: magic `const`s. This code base has inlined constants increasingly being spread out. Please consider taking a seperate PR/pass centralizing and giving meaningful names. <3

michaelcfanning: Will do.

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
michaelcfanning: 100% code coverage! :)

Fuzzing. Oh yes, we should do that. For .NET fuzzing is very useful for hardening reliability of API (as well as gleaning the typical security value).

PR #36, opened 2024-05-31
Bundling Rust DLL into Nuget

suvamM:
This PR makes the following changes:

1. Builds the Rust FFI project for x86 and x86-64 Windows.
2. Bundles the x86 and x86-64 Rust dlls into the Nuget. No Nuspec file is required.
3. Creates a sample project that ingests the locally created Nuget, and performs some basic operation to test the packaging. This sample project _failed_ prior to the Rust dll packaging in the Nuget, and passes after the dll was correctly packaged.


refs/win-x64/microsoft_security_utilities_core.dll
rwoll, line 1: Please remove these files from the repo and add a gitignore for `*.dll` and `refs/` if not already present.

These should be built during pipeline and not checked in binaries.

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
```diff
@@ -54,4 +54,11 @@
     </PackageReference>
   </ItemGroup>
 
+	<ItemGroup>
```
rwoll: nit: indentation is off from the other code in this file.

BuildAndTest.cmd
```diff
@@ -19,18 +19,20 @@ if "%ERRORLEVEL%" NEQ "0" (
...
   echo "security_utilities_rust_ffi build failed..."
   exit /b %ERRORLEVEL%
 )
 popd
 
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.dll .\refs
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.pdb .\refs
+xcopy /Y .\src\security_utilities_rust_ffi\target\i686-pc-windows-msvc\release\microsoft_security_utilities_core.dll .\refs\win-x86
+xcopy /Y .\src\security_utilities_rust_ffi\target\i686-pc-windows-msvc\release\microsoft_security_utilities_core.pdb .\refs\win-x86
+xcopy /Y .\src\security_utilities_rust_ffi\target\x86_64-pc-windows-msvc\release\microsoft_security_utilities_core.dll .\refs\win-x64
```
rwoll: Is this overwriting i686 dll?

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Nuget.config
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rwoll, line 1: Please see example for Spmi. This likely needs more bits to achieve the desired behavior.

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/SecurityUtilitiesApiUtilizationExample.csproj
```diff
@@ -0,0 +1,14 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="1.4.22-gda2bd0f88d" />
```
rwoll: Use `*-*` here since this specific version, if everything is configured correctly, won't be the nupkg version you're trying to test.

rwoll (changes requested):

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Nuget.config
suvamM: @rwoll Didn't follow this. All we want to test is the single Nuget `Microsoft.Security.Utilities.Core`. What behaviors are missed here?

BuildAndTest.cmd
suvamM: @rwoll The x86 target in Rust is called i686. This is copying the Rust compiled i686 binaries into the 32-bit runtime folder in `refs`.

rwoll: Oops! Numbers! Thx!

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+                {
+                    for (int i = 0; i < iterations; i++)
+                    {
+                        foreach (string signature in identifiable.Signatures!)
+                        {
+                            string key = IdentifiableSecrets.GenerateStandardBase64Key(seed,
+                                                                                     identifiable.KeyLength,
+                                                                                     signature);
+
+                            int found = masker.DetectSecrets(key).Count();
```
michaelcfanning: >int found = masker.DetectSecrets(key).Count() [](http://example.com/codeflow?start=28&length=45)

you don't have any validation of function in your test.

neither do you return or check a success code. Maybe a non-zero result is set for an unhandled exception in your script? Will go look.

My specific suggestion is that you verify we actually find secrets here, a non-zero count is probably sufficient.

Please don't block on this note, let's get this in and you can refine.

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+    internal class Program
+    {
+        static void Main(string[] args)
+        {
+            int iterations = 10;
+
+            var masker = new IdentifiableScan(WellKnownRegexPatterns.HighConfidenceMicrosoftSecurityModels,
+                                              generateCorrelatingIds: false);
+
+            foreach (var pattern in WellKnownRegexPatterns.HighConfidenceSecurityModels)
```
michaelcfanning, line 16: >HighConfidenceSecurityModels [](http://example.com/codeflow?start=59&length=28)

why would you limit yourself here? it seems to me you'd prefer more testing rather than less.

BuildAndTest.cmd
```diff
@@ -19,18 +19,20 @@ if "%ERRORLEVEL%" NEQ "0" (
 )
 popd
 
-
 pushd .\src\security_utilities_rust_ffi\
 call cargo clean --release
-call cargo build --release
+call cargo build --release --target x86_64-pc-windows-msvc
+call cargo build --release --target i686-pc-windows-msvc
```
michaelcfanning, line 29: >windows [](http://example.com/codeflow?start=44&length=7)

why no linux support? is this coming later?

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
michaelcfanning: you are set to four spaces. one of my goofs is i typically set the IDE to 2 spaces for XML due to its tendency towards nesting (and therefore indentation)

BuildAndTest.cmd
```diff
@@ -19,18 +19,20 @@ if "%ERRORLEVEL%" NEQ "0" (
 )
 popd
 
-
 pushd .\src\security_utilities_rust_ffi\
 call cargo clean --release
-call cargo build --release
+call cargo build --release --target x86_64-pc-windows-msvc
```
michaelcfanning, line 24: >build [](http://example.com/codeflow?start=11&length=5)

you don't test the error code for the first build. it could fail, the second build succeeds and sets errorlevel to 0

BuildAndTest.cmd
```diff
@@ -19,18 +19,20 @@ if "%ERRORLEVEL%" NEQ "0" (
...
 if "%ERRORLEVEL%" NEQ "0" (
   echo "security_utilities_rust_ffi build failed..."
   exit /b %ERRORLEVEL%
 )
 popd
 
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.dll .\refs
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.pdb .\refs
+xcopy /Y .\src\security_utilities_rust_ffi\target\i686-pc-windows-msvc\release\microsoft_security_utilities_core.dll .\refs\win-x86
+xcopy /Y .\src\security_utilities_rust_ffi\target\i686-pc-windows-msvc\release\microsoft_security_utilities_core.pdb .\refs\win-x86
```
michaelcfanning: >pdb [](http://example.com/codeflow?start=113&length=3)

i don't see your checked in pdbs. pdbs won't get added automatically via git.exe you need to add the --force arg to get er done.

BuildAndTest.cmd
```diff
@@ -19,18 +19,20 @@ if "%ERRORLEVEL%" NEQ "0" (
...
+call cargo build --release --target x86_64-pc-windows-msvc
+call cargo build --release --target i686-pc-windows-msvc
 if "%ERRORLEVEL%" NEQ "0" (
   echo "security_utilities_rust_ffi build failed..."
   exit /b %ERRORLEVEL%
 )
 popd
 
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.dll .\refs
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.pdb .\refs
```
michaelcfanning, line 33: >microsoft_security_utilities_core [](http://example.com/codeflow?start=58&length=33)

this change needs to delete the checked in files

michaelcfanning:
> call cargo clean --release

this script is in an inconsistent form. you clean for the general release build (so we can trust testing). you don't clean for the actual product build (so we may be subject to caching issues).

I suggest you remove the clean steps and make them required by passing an argument (that our pipeline should invoke).

this way, local dev build is fast. but devs can opt into a full clean to be 100% sure of things.

---
Refers to: BuildAndTest.cmd:5 in be184c7. [](commit_id = be184c752cdb722ce24c90ea8c88cef631b9a2d1, deletion_comment = False)

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Nuget.config
rwoll: The nuget.config you have does not override enough locations to force it to use the local packages. Here's an updated version which has the added benefit that you can easily inspect/delete `globalPackagesFolderTest` to see which exact nuget packages you're pulling in. Be sure to add a `.gitignore` entry for `globalPackagesFolderTest/`

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <config>
    <add key="globalPackagesFolder" value="globalPackagesFolderTest" />
  </config>
  <fallbackPackageFolders>
    <add key="LocalBuild" value="..\..\..\bld\nupkg\AnyCPU_Release\" />
  </fallbackPackageFolders>
  <packageSources>
    <clear />
    <add key="LocalBuild" value="..\..\..\bld\nupkg\AnyCPU_Release\" />
    <add key="PublicNugetOrg" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
  <packageSourceMapping>
    <packageSource key="LocalBuild">
	  <package pattern="Microsoft.Security.Utilities.*" />
      <package pattern="Sarif.*" />
    </packageSource>
    <packageSource key="PublicNugetOrg">
      <package pattern="*" />
    </packageSource>
  </packageSourceMapping>
</configuration>
```

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/SecurityUtilitiesApiUtilizationExample.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
```
rwoll: does everything still work if you target `net462`?

rwoll: Also be sure to test x64 v. x86 to confirm that the expected DLLs actually load.

BuildAndTest.cmd
rwoll: PDBs and DLLs should not be checked into the repo except for extenuating circumstances.

The PDBs and DLLs shipped in the package should be built from source, not some random copy from the repo. If they are in the repo checkout, you risk having these checked in copies accidentally being used in the tests and/or copied+used in the distributed nupkg.

In fact, as of right now, this PR is running the wrong/outdated DLL in `src\Tests.Microsoft.Security.Utilities.Core\Tests.Microsoft.Security.Utilities.Core.csproj`.

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Program.cs
suvamM: @michaelcfanning I specifically don't want to perform any functional test here. Shouldn't the objective here be limited to testing whether the nuget loads and operates on relevant frameworks and platforms?

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
suvamM: Fixed.

michaelcfanning (dismissed):
:shipit:

michaelcfanning:
Is this change functional? If so can we please get it in then work on tail tasks?

refs/win-x64/microsoft_security_utilities_core.dll
suvamM: Fixed.

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Nuget.config
suvamM: Fixed.

BuildAndTest.cmd
suvamM: @michaelcfanning That'll be a follow-on to this PR

BuildAndTest.cmd
suvamM: Good catch. Fixed.

BuildAndTest.cmd
suvamM: Fixed.

.gitignore
```diff
@@ -364,3 +365,9 @@ Cargo.lock
 
 # MSVC Windows builds of rustc generate these, which store debugging information
 *.pdb
+
+# Ignore the globalPackagesFolderTest
+**/globalPackagesFolderTest/
```
rwoll, line 370:
```suggestion
globalPackagesFolderTest/
```

nit: can be simplified per the docs:

> For example, a pattern doc/frotz/ matches doc/frotz directory, but not a/doc/frotz directory; however frotz/ matches frotz and a/frotz that is a directory (all paths are relative from the .gitignore file).

Source: https://git-scm.com/docs/gitignore

BuildAndTest.cmd
```diff
@@ -19,18 +19,25 @@ if "%ERRORLEVEL%" NEQ "0" (
...
+if "%ERRORLEVEL%" NEQ "0" (
+  echo "security_utilities_rust_ffi build failed for i686-pc-windows-msvc target..."
+  exit /b %ERRORLEVEL%
+)
+
 popd
 
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.dll .\refs
-xcopy /Y .\src\security_utilities_rust_ffi\target\release\microsoft_security_utilities_core.pdb .\refs
+xcopy /Y .\src\security_utilities_rust_ffi\target\i686-pc-windows-msvc\release\microsoft_security_utilities_core.dll .\refs\win-x86\
```
rwoll, line 37: when you add support for linux and darwin, please refactor this script so this block as well as the cargo build bits are not repeated a bunch.

There's a lot of numbers for the eyes to keep track of, so DRY'ing it up will help ensure there's not something subtly out of sync between these various sections.

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
```diff
@@ -54,4 +54,11 @@
     </PackageReference>
   </ItemGroup>
 
+  <ItemGroup>
+	<None Include="..\..\refs\win-x64\*.dll" Pack="true" PackagePath="runtimes\win-x64\native" />
```
rwoll, line 58: This should move into `Build.props` to be colocated with the dev-time copying.

rwoll: Internal builds use Build.props but not the crpoj.

rwoll (approved):

PR #37, opened 2024-05-31
Add long form option to request the full HIS v2 key. Expand reporting…

michaelcfanning:
… for negative cases.

- BRK: Add `longForm` argument to `IdentifiableSecrets.GenerateCommonAnnotatedKey`, to produce the optional full 64-byte form (which includes the full 4-byte Marvin checksum).

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -11,6 +11,7 @@
 - FNS => Flase negative reduction in static analysis.
 
 # UNRELEASED
+- BRK: Add `longForm` argument to `IdentifiableSecrets.GenerateCommonAnnotatedKey`, to produce the optional full 64-byte form (which includes the full 4-byte Marvin checksum).
 - BRK: Coalesce `AadClientAppIdentifiableCredentialsCurrent` and `AadClientAppIdentifiableCredentialsPrevious` into a single `AadClientAppIdentifiableCredentials` check.
 - BRK: Rename `IIdentifiableKey.SniffLiterals` to `IIdentifiableKey.Signatures` to precisely reflect their purpose to signify fixed signatures in keys.
 - BUG: Update `IdentifiableScan` to post-process finds (e.g., with checksum validation) to eliminate false positives.
--- a/src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
@@ -19,12 +19,19 @@ namespace Microsoft.Security.Utilities
             string base64EncodedSignature = key.Substring(76, 4);
             bool isDerived = key[DerivedKeyCharacterOffset] == 'D';
 
-            if (key.Length != 88 && key.Length != 84)
+            if (key.Length != IdentifiableSecrets.StandardCommonAnnotatedKeySize &&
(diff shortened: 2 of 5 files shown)


src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
```diff
@@ -19,12 +19,19 @@ public static bool TryCreate(string key, out CommonAnnotatedKey secret)
...
                 return false;
             }
 
-            if (key.Length == 84)
+            bool longForm = key.Length == IdentifiableSecrets.LongFormCommonAnnotatedKeySize;
+
+            // This code path is intended to ensure that common annotated security keys are
+            // highly backwards compatible with the older identifiable keys format. This
+            // should only entail providing the missing compute checksum byte of the key if
+            // it is absent. The validation performed below is a fairly expensive operation.
```
michaelcfanning, line 33: >expensive [](http://example.com/codeflow?start=72&length=9)

we actually might want a knob in the future to prevent the heavyweight validation.

src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
```diff
@@ -70,7 +77,63 @@ private CommonAnnotatedKey(byte[] bytes)
             this.base64Key = Convert.ToBase64String(this.bytes);
         }
 
-        public const int DerivedKeyCharacterOffset = 57;
+        /// <summary>
+        /// The offset to the encoded standard fixed signature ('JQQJ99' or 'JQQJ9D').
+        /// </summary>
+        public const int StandardFixedSignatureOffset = 52;
```
michaelcfanning, line 83: >StandardFixedSignatureOffset [](http://example.com/codeflow?start=25&length=28)

this is Ross' `death to magic numbers` PR

rwoll: Thank you! Thank you! Thank you! This goes a long way for readability and will help prevent the mistake from previous PRs.

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,7 @@
 - FNS => Flase negative reduction in static analysis.
 
 # UNRELEASED
+- BRK: Add `longForm` argument to `IdentifiableSecrets.GenerateCommonAnnotatedKey`, to produce the optional full 64-byte form (which includes the full 4-byte Marvin checksum).
```
rwoll, line 14: Which part is `BRK`ing? Looks like `longForm` is added with a default for `false`, so it shouldn't impact existing users.

michaelcfanning: This is subtle. If you happened to explicitly provide the `testChar` call (without using a named parameter) the introduction of a new parameter with a default value could cause a compiler breaking change.

src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
michaelcfanning: Yes indeed! Thanks for advocating for the right things. The newly added unit tests, by the way, identified the problem we both observed via code inspection.

rwoll (approved):

PR #38, opened 2024-06-02
Add detections for loose sas secret and generic JWT exposures.

michaelcfanning:
- RUL: Add `SEC101/051.LooseSasSecret` detection.
- RUL: Add `SEC101/528.GenericJwt` detection.


src/Microsoft.Security.Utilities.Core/HighConfidenceMicrosoftSecurityModels/SEC101_101_AzureClientAppLegacyCredential32.cs
```diff
@@ -47,6 +47,12 @@ public override IEnumerable<string> GenerateTestExamples()
             }
         }
 
+        public override IEnumerable<string> GenerateFalsePositiveExamples()
+        {
+            yield return DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffffffZ");
```
michaelcfanning, line 57: >DateTime [](http://example.com/codeflow?start=25&length=8)

Someone out there knows of false positives that currently provoke this rule. This one isn't a good example! it doesn't even meet regex length requirements. I'm soliciting an actual repro now. #Resolved

src/Microsoft.Security.Utilities.Core/ExtensionMethods.cs
```diff
@@ -12,5 +12,9 @@ public static ISet<string> ToSet(this string value)
         {
             return new HashSet<string> { value };
         }
+        public static ISet<string> ToSet(this IEnumerable<string> value)
```
rwoll, line 15: nit: will `public` visibility of these extension methods litter customers of `Microsoft.Security.Utilities` namespace?

I suggest changing to `internal` or moving them all to their own namespace so customers can decide whether or not to include them `using Microsoft.Security.Utilities.Extensions`. #Resolved

src/Microsoft.Security.Utilities.Core/HighConfidenceMicrosoftSecurityModels/SEC101_051_LooseSasSecret.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Web;
+
+namespace Microsoft.Security.Utilities
+{
+    public class LooseSasSecret : RegexPattern
+    {
+        public LooseSasSecret()
+        {
+            Id = "SEC101/051";
```
rwoll: This id does not look consistent with existing data. SEC101/051 is stripe. #Resolved

src/Microsoft.Security.Utilities.Core/HighConfidenceMicrosoftSecurityModels/SEC101_051_LooseSasSecret.cs
```diff
@@ -0,0 +1,27 @@
...
+            Name = nameof(LooseSasSecret);
+            DetectionMetadata = DetectionMetadata.HighEntropy;
+            Pattern = @$"(?i)(?:^|[?;&])(?:dsas_secret|sig)=(?<refine>[0-9a-z\/+%]{{43,}}(?:=|%3d))";
```
rwoll: Are Sas's guaranteed to have the padding? If not, please refine and add a test case below. #Resolved

src/Microsoft.Security.Utilities.Core/LowConfidenceClassifiers/SEC101_528_GenericJwt.cs
```diff
@@ -0,0 +1,65 @@
...
+            else
+            {
+                // Padding needed
+                return new string('=', 4 - remainder);
+            }
+        }
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            yield return $"eyAidHlwIiA6ICJKV1QiLCAiYWxnIiA6ICJIUzI1NiIgfQ.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
```
rwoll, line 66: The other locations you generate examples. Why inline these?

I don't necessarily object — concrete examples are important, too. #Pending

rwoll, line 66: Do these cases have a title? What are you trying to test with each of the three of these? #Resolved

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
```diff
@@ -43,6 +43,10 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="System.Web" />
```
rwoll: Does this warrant a release note? #Resolved

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -49,6 +49,8 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
 
     public static IEnumerable<RegexPattern> LowConfidencePotentialSecurityKeys { get; } = new RegexPattern[]
     {
+        new GenericJwt(),
```
rwoll, line 52: aside: this patten smells a bit. why do these definitions require instantiation? Couldn't they all be static functions/data? is there any state within them?

C# 8 FTW: https://stackoverflow.com/questions/9415257/how-can-i-implement-static-methods-on-an-interface/60083436#60083436 #Pending

rwoll (changes requested):

src/Microsoft.Security.Utilities.Core/ExtensionMethods.cs
michaelcfanning: Good points. So the way I handle this typically is to encode the assembly name somehow in the extension methods class. And so for SARIF, for example, the Driver class is called DriverExtensionMethods. This keeps everything from colliding. Now for SARIF, the core SARIF dll represents the root namespace and therefore obtains the basic `ExtensionMethods` name. Following that pattern, this class s/be `CoreExtensionMethods` will do that.

Note that the class itself is already internal, so altering the method level accessibility won't have any impact.

src/Microsoft.Security.Utilities.Core/HighConfidenceMicrosoftSecurityModels/SEC101_051_LooseSasSecret.cs
michaelcfanning: Good catch, thank you.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
michaelcfanning: The issue is in the classes of secret that implement `IIdentifiableKey`. Now we could still use a singleton pattern for all of this code. Let's talk offline about it (simply meaning outside of this PR context)

src/Microsoft.Security.Utilities.Core/LowConfidenceClassifiers/SEC101_528_GenericJwt.cs
michaelcfanning: added a comment.

michaelcfanning: just the 'waves on the beach' model. later in a quieter moment will add auto-generators. i didn't do it here simply because it will require hand-authoring some code, as I don't want to introduce dependencies for generating JWTs here. One weakness of implementing the test example generation directly in the product logic is that it constrains the dependencies we can (or should) add. this suggests a better pattern might be to actually have the test code layer this capability onto the product.

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
michaelcfanning: OK, so my first thought was no, simply because this dependency is always available in the CLR at runtime. And so we don't expect to be subject to any of the usual breaking issues that warrant meticulous documentation of new dependencies.

I'm glad you cited this, though, because this assembly is very free of explicit references and this one was pulled in simply to generate a single test pattern (which I've now replaced with a string constant).

docs/ReleaseHistory.md
```diff
@@ -11,6 +12,11 @@
 - FNS => Flase negative reduction in static analysis.
 
 # UNRELEASED
+- RUL: Add `SEC101/051.LooseSasSecret` detection.
+- RUL: Add `SEC101/528.GenericJwt` detection.
+- BRK: Rename `WellknownPatterns.HighConfidenceSecurityModels` to `WellknownPatterns.PreciselyClassifiedSecurityKeys`.
+- BRK: Rename `WellknownPatterns.LowConfidencePotentialSecurityKeys` to `WellknownPatterns.UnclassifiedPotentialSecurityKeys`.
```
suvamM, line 18: The `HighConfidence` and `LowConfidence` nomenclature was better. It conveys some information about the detection capability. For `Unclassified`, it's less clear---why is it unclassified?

This is a not a blocking comment.

src/Microsoft.Security.Utilities.Core/LowConfidenceClassifiers/SEC101_060_LooseSasSecret.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Collections.Generic;
+using System.Web;
+
+namespace Microsoft.Security.Utilities
+{
+    public class LooseSasSecret : RegexPattern
+    {
+        public LooseSasSecret()
+        {
+            Id = "SEC101/060";
```
suvamM, line 13: The Release Notes sets the rule ID to `SEC101/051`. This is a bug. What's the correct rule ID? #Resolved

src/Microsoft.Security.Utilities.Core/LowConfidenceClassifiers/SEC101_060_LooseSasSecret.cs
```diff
@@ -0,0 +1,47 @@
...
+            Name = nameof(LooseSasSecret);
+            DetectionMetadata = DetectionMetadata.HighEntropy;
+            Pattern = @$"(?i)(?:^|[?;&])(?:dsas_secret|sig)=(?<refine>[0-9a-z\/+%]{{43,129}}(?:=|%3d))";
+            Signatures = new HashSet<string>(new[] { "sig=", "ret=" });
+        }
+
+
+        // HttpUtility.UrlEncode(new string('/', 43));
+        private const string FortyThreeBase64EncodedSlashes = "%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f%2f";
```
suvamM, line 22: This is hard to maintain---what if someone mistakenly nukes a %2f? This only seems to be used in generating examples. Can't we add a helper method that creates 43 %2f on the fly and returns it? Is this string on a perf sensitive path?

docs/ReleaseHistory.md
```diff
@@ -11,6 +12,11 @@
 - FNS => Flase negative reduction in static analysis.
 
 # UNRELEASED
+- RUL: Add `SEC101/051.LooseSasSecret` detection.
```
suvamM: should be 060? #Resolved

suvamM (changes requested):

docs/ReleaseHistory.md
michaelcfanning: Thanks! Fixed.

docs/ReleaseHistory.md
michaelcfanning: I don't agree. In fact, the literal accuracy/confidence of these checks varies and in some cases, a rule that doesn't actually classify a finding may be fairly high confidence. In practice, our generic JWT rule doesn't fire many false positives for example.

So what I've done is separate rules by those that attempt to specifically classify a finding (this is an event hub key) vs. those that don't (this is an unclassified JWT).

These designations say something about volume of finds and the literal utility of findings, which is why I separated them.

Categorization is hard, so I'm definitely interested in the ongoing conversation here. What does 'High' vs. 'Low' confidence mean to you? What criterion should we use to decide where a rule would be placed given these two buckets? This is where I was struggling.

src/Microsoft.Security.Utilities.Core/LowConfidenceClassifiers/SEC101_060_LooseSasSecret.cs
michaelcfanning: `060` thanks for noticing.

src/Microsoft.Security.Utilities.Core/LowConfidenceClassifiers/SEC101_060_LooseSasSecret.cs
michaelcfanning: I wouldn't provide a helper because this is a const value for a contrived case. There is a code comment that describes how it was produced. 

If someone nukes a %2f that still matches the pattern. Why would a use go tinker with this exactly? I assume they will put that change through code review? If there was a helper that, for example, constructed this string, what if someone mistakenly changed that code? Just trying to understand what you're defending against. :)

suvamM (approved):

rwoll (approved):

GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
```diff
@@ -0,0 +1,404 @@
...
+    ],
+    "DetectionMetadata": "HighEntropy, FixedSignature"
+  },
+  {
+    "Pattern": "^(?i)[a-z0-9.=\\-:[_@\\/*\\]+?]{32}$",
+    "Id": "SEC101/101",
+    "Name": "AadClientAppLegacyCredentials",
+    "Signatures": null,
+    "DetectionMetadata": "ObsoleteFormat, HighEntropy"
+  },
```
evelyn-ys, line 335: Per my test, a timestamp like `2024-03-07T02:50:56.464790+00:00` would match this pattern as it's a simple string containing only numbers, letter `T` and symbols `-`, `:`, `.` and `+` while the length exactly equals 32

docs/ReleaseHistory.md
evelyn-ys: I'd prefer that if a rule fires some false positives, then we could put it in Low confidence. If no false positive feedback received or identified, then we can leave it as High confidence. So basically all rules are high confidence at the beginning, we just need a new file for low confidence when we find false positives from Classified patterns or Unclassified patterns

PR #42, opened 2024-06-04
Fix identifiable scan behaviors and add hybrid scan capability.

michaelcfanning:
- BUG: Bring `IdentifiableScan` into precise equivalence with other maskers, e.g., `Detection.RedactionToken` is now in alignment.
- NEW: Provide hybrid capability to run high-performance detections in `IdentifiableScan` and fall back to other masker as required.

This change extends our unit testing to ensure that the high performance Rust library is exactly equivalent in behavior to the other libraries. I noticed a discrepancy in some ad hoc testing in redaction token production. Testing also closed a few other holes.

Opportunistically, to enable full testing I updated the high perf scanner to create a back-up masker for unimplemented rules. This allows the high performance masker to be initialized with arbitrary patterns. If it can assume responsibilities for detection, it does so, otherwise it uses a fallback masker.


src/Microsoft.Security.Utilities.Core/IdentifiableScan.cs
```diff
@@ -65,27 +65,64 @@ public class IdentifiableScan : ISecretMasker, IDisposable
...
-    private IntPtr scan;
-    private readonly Dictionary<string, ISet<string>> idToSignaturesMap;
     private readonly Dictionary<string, IList<RegexPattern>> signatureToPatternsMap;
+    private readonly Dictionary<string, ISet<string>> idToSignaturesMap;
+    private SecretMasker backupSecretMasker;
     private List<string> orderedIds;
-    
+    private IntPtr scan;
+
+    private static readonly ISet<string> HighPerformanceEnabledSignatures = new HashSet<string>(new string[]
```
michaelcfanning, line 74: >HighPerformanceEnabledSignatures [](http://example.com/codeflow?start=41&length=32)

We really want this list to either be discoverable by calling Rust *or* we should pass rule definitions to Rust, allowing us to maintain rulesets here in C#. TBD.

src/Microsoft.Security.Utilities.Core/IdentifiableScan.cs
```diff
@@ -65,27 +65,64 @@ public class IdentifiableScan : ISecretMasker, IDisposable
...
         this.signatureToPatternsMap = new Dictionary<string, IList<RegexPattern>>();
+        this.idToSignaturesMap = new Dictionary<string, ISet<string>>();
+        this.generateCorrelatingIds = generateCorrelatingIds;
         this.orderedIds = new List<string>();
 
         foreach (RegexPattern pattern in regexPatterns)
         {
             if (pattern.Signatures == null)
             {
+                this.backupSecretMasker ??= new SecretMasker(regexSecrets: null,
```
michaelcfanning: Any regex with no signature is not a candidate for the high-perf detection, so we simply slam it into the back-up engine.

src/Microsoft.Security.Utilities.Core/IdentifiableScan.cs
```diff
@@ -249,6 +286,14 @@ public IEnumerable<Detection> DetectSecrets(string input)
         {
             yield return detection;
         }
+
+        if (this.backupSecretMasker != null)
```
michaelcfanning, line 295: >backupSecretMasker [](http://example.com/codeflow?start=17&length=18)

currently, we only apply the back-up masker when scanning textual data, received as a populated string.

to make everything stream friendly would require getting into the IRegexPattern interface and some other places.

src/Microsoft.Security.Utilities.Core/IdentifiableScan.cs
```diff
@@ -315,30 +360,37 @@ public IEnumerable<Detection> DetectSecrets(Stream file)
...
+                                        : RegexPattern.FallbackRedactionToken;
+
                                     yield return new Detection
                                     {
-                                        Id = result.Item1,
+                                        Id = preciseId,
                                         Name = result.Item2,
                                         Start = (int)start,
                                         Length = (int)textLength,
+                                        Metadata = DetectionMetadata.HighEntropy,
```
michaelcfanning, line 397: >HighEntropy [](http://example.com/codeflow?start=69&length=11)

All rules handled by the high perf scanner are high entropy. Mostly these checks are identifiable, we need to update this metadata to be more of a first class property (bundled with signatures, passed to Rust).

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/SecurityUtilitiesApiUtilizationExample.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-	  <TargetFrameworks>net462;net6.0</TargetFrameworks>
+	  <TargetFrameworks>net6.0</TargetFrameworks>
```
suvamM: Why are we dropping the check for `net462`? This will cause the `BuildAndTest.cmd` to crash, since it explicitly invokes the binaries in `net462`.

suvamM (dismissed):

suvamM (approved):

PR #45, opened 2024-06-05
Add SEC101_061_LooseOAuth2BearerToken and notion of confidence

michaelcfanning:
- RUL: Add `SEC101/061.LooseOAuth2BearerToken` detection.
- NEW: Add `DetectionMetadata.LowConfidence` and `Detection.MediumConfidence` designations.

This change first adds a detection requested by a partner, for detecting and redacting a loose OAuth2.0 bearer token. The data stored in this header location appears to itself have different patterns and conventions (for example, a simple base64 encoded token or a CBOR web token).

The corresponding rule doesn't make an effort to classify these or try to determine whether the provided value is recognizable, e.g., as a 32-byte base64-encoded value.

That leads me to mark is a 'low confidence'. i.e., we expect to redact test/other non-sensitive data as well as actual secrets with this pattern.

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -11,6 +11,10 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
 
+# 1.5 PLEASE START TO VERSION MINOR NUMBER FOR BREAKING CHANGES
+- RUL: Add `SEC101/061.LooseOAuth2BearerToken` detection.
+- NEW: Add `DetectionMetadata.LowConfidence` and `Detection.MediumConfidence` designations.
+
 # 1.4.25 - 06/04/2024
 - BUG: Bring `IdentifiableScan` into precise equivalence with other maskers, e.g., `Detection.RedactionToken` is now in alignment.
 - NEW: Provide hybrid capability to run high-performance detections in `IdentifiableScan` and fall back to other masker as required.
--- a/src/Microsoft.Security.Utilities.Core/UnclassifiedPotentialSecurityKeys/SEC101_061_LooseOAuth2BearerToken.cs
+++ b/src/Microsoft.Security.Utilities.Core/UnclassifiedPotentialSecurityKeys/SEC101_061_LooseOAuth2BearerToken.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class OAuth2BearerToken : RegexPattern
+    {
+        public OAuth2BearerToken()
+        {
+            Id = "SEC101/061";
+            Name = nameof(OAuth2BearerToken);
+            DetectionMetadata = DetectionMetadata.LowConfidence;
+
+            // https://datatracker.ietf.org/doc/html/rfc6750#section-2.1
+            Pattern = $"(?i)authorization:(\\s|%20)bearer(\\s|%20)<?<refine>[0-9a-z][{WellKnownRegexPatterns.UrlUnreserved}+/=]*)[^[{WellKnownRegexPatterns.UrlUnreserved}+/=]";
+            Signatures = new HashSet<string>(new[] { "sig=", "ret=" });
+        }
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            yield return $"Authorization: bearer notasecret==";
+            yield return $"authorization:%20bearer%20secretplacholder";
+        }
+    }
+}
--- a/GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json
+++ b/GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json
@@ -8,7 +8,7 @@
       "eyAi",
       "ewog"
     ],
-    "DetectionMetadata": "HighEntropy"
+    "DetectionMetadata": "HighEntropy, MediumConfidence"
   },
   {
     "Pattern": "(?i)(?:^|[?;&])(?:dsas_secret|sig)=(?<refine>[0-9a-z\\/+%]{43,129}(?:=|%3d))",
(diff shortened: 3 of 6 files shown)


michaelcfanning:
>     "Name": "Unclassified32ByteBase64String",

So what do we designate these checks in terms of confidence? On the one hand, we are pretty sure we have a 32-byte base64 string. :) On the other, we don't have much confidence we've identified a secret.

It all just shows how hard it is to categorize these checks. :)

I haven't provided any confidence at all yet, because these checks really only describe the literal format of the detected token, there is no attempt to categorize as a secret whatsoever. We should think about and make a call at some point though.

btw - the application of these patterns is to sample/review data for missed detections.

---
Refers to: GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json:26 in 946a003. [](commit_id = 946a00378b46d06e03517860d56e3a9b4080a56d, deletion_comment = False)

michaelcfanning:
>     "Id": "SEC000/000",

So what do we designate these checks in terms of confidence? On the one hand, we are pretty sure we have a 32-byte base64 string. :) On the other, we don't have much confidence we've identified a secret.

It all just shows how hard it is to categorize these checks. :)

I haven't provided any confidence at all yet, because these checks really only describe the literal format of the detected token, there is no attempt to categorize as a secret whatsoever. We should think about and make a call at some point though.

btw - the application of these patterns is to sample/review data for missed detections.

---
Refers to: GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json:25 in 946a003. [](commit_id = 946a00378b46d06e03517860d56e3a9b4080a56d, deletion_comment = False)

michaelcfanning:
>     "Id": "SEC000/000",

So what do we designate these checks in terms of confidence? On the one hand, we are pretty sure we have a 32-byte base64 string. :) On the other, we don't have much confidence we've identified a secret.

It all just shows how hard it is to categorize these checks. 

I haven't provided any confidence at all yet, because these checks really only describe the literal format of the detected token, there is no attempt to categorize as a secret whatsoever. We should think about and make a call at some point though.

btw - the application of these patterns is to sample or review data for missed detections.

---
Refers to: GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json:25 in 946a003. [](commit_id = 946a00378b46d06e03517860d56e3a9b4080a56d, deletion_comment = False)

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,10 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
 
+# 1.5 PLEASE START TO VERSION MINOR NUMBER FOR BREAKING CHANGES
+- RUL: Add `SEC101/061.LooseOAuth2BearerToken` detection.
```
michaelcfanning, line 15: >061 [](http://example.com/codeflow?start=19&length=3)

is this id the next one? we should maintain a primary list of assigned things.

src/Microsoft.Security.Utilities.Core/UnclassifiedPotentialSecurityKeys/SEC101_061_LooseOAuth2BearerToken.cs
```diff
@@ -0,0 +1,27 @@
...
+        public OAuth2BearerToken()
+        {
+            Id = "SEC101/061";
+            Name = nameof(OAuth2BearerToken);
+            DetectionMetadata = DetectionMetadata.LowConfidence;
+
+            // https://datatracker.ietf.org/doc/html/rfc6750#section-2.1
+            Pattern = $"(?i)authorization:(\\s|%20)bearer(\\s|%20)<?<refine>[0-9a-z][{WellKnownRegexPatterns.UrlUnreserved}+/=]*)[^[{WellKnownRegexPatterns.UrlUnreserved}+/=]";
+            Signatures = new HashSet<string>(new[] { "sig=", "ret=" });
+        }
```
evelyn-ys, line 19: This is added in .NET code but seems not added to `UnclassifiedPotentialSecurityKeys.json`

evelyn-ys: I'm thinking there could be some cases that BearerToken is used without the prefix `authorization:`

src/Microsoft.Security.Utilities.Core/UnclassifiedPotentialSecurityKeys/SEC101_061_LooseOAuth2BearerToken.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class OAuth2BearerToken : RegexPattern
```
suvamM, line 8: This needs to be added to `WellKnownRegexPatterns.UnclassifiedPotentialSecurityKeys`. Pushing this point a bit more, perhaps it would be nice to autogenerate the list?

src/Microsoft.Security.Utilities.Core/UnclassifiedPotentialSecurityKeys/SEC101_061_LooseOAuth2BearerToken.cs
```diff
@@ -0,0 +1,27 @@
...
+            Id = "SEC101/061";
+            Name = nameof(OAuth2BearerToken);
+            DetectionMetadata = DetectionMetadata.LowConfidence;
+
+            // https://datatracker.ietf.org/doc/html/rfc6750#section-2.1
+            Pattern = $"(?i)authorization:(\\s|%20)bearer(\\s|%20)<?<refine>[0-9a-z][{WellKnownRegexPatterns.UrlUnreserved}+/=]*)[^[{WellKnownRegexPatterns.UrlUnreserved}+/=]";
+            Signatures = new HashSet<string>(new[] { "sig=", "ret=" });
+        }
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
```
suvamM, line 21: What's the test for this newly added rule?

suvamM (approved):

PR #46, opened 2024-06-14
Fixing SEC101/200 GenerateTruePositiveExamples test case

suvamM:
This PR
1. Adds `CommonAnnotatedSecurityKey` to the high confidence security models set.
2. Fixes the test in `GenerateTruePositiveExamples` for `CommonAnnotatedSecurityKey`. The logic was completely broken.
3. Added release note entry for the same.

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -9,11 +9,13 @@
 - NEW => New API or feature.
 - PRF => Performance work.
 - FPS => False positive reduction in static analysis.
-- FNS => Flase negative reduction in static analysis.
+- FNS => False negative reduction in static analysis.
 
 # 1.5 PLEASE START TO VERSION MINOR NUMBER FOR BREAKING CHANGES
 - RUL: Add `SEC101/061.LooseOAuth2BearerToken` detection.
 - DEP: Added support for net451 in `Microsoft.Security.Utilities.Core` for backward compatibility.
+- BUG: Fix the logic in `CommonAnnotatedSecurityKey.GenerateTruePositiveExamples()` to handle invalid test key characters, and to properly break out of the testing loop.
+- FNS: Added `SEC101/200.CommonAnnotatedSecurityKey` to `WellKnownPatterns.HighConfidenceMicrosoftSecurityModels`.
 - NEW: Add `DetectionMetadata.LowConfidence` and `Detection.MediumConfidence` designations.
 
 # 1.4.25 - 06/04/2024
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
@@ -22,25 +22,40 @@ namespace Microsoft.Security.Utilities
             int count = 0;
             int attempts = 0;
 
-            while(true)
+            foreach (bool longForm in new[] { true, false })
             {
-                foreach (bool longForm in new[] { true, false })
+                while (true) 
                 {
                     char testChar = (char)('a' + attempts++);
-                    string example = IdentifiableSecrets.GenerateCommonAnnotatedTestKey(IdentifiableSecrets.VersionTwoChecksumSeed,
-                                                                                        "TEST",
-                                                                                        customerManagedKey: true,
-                                                                                        platformReserved: null,
-                                                                                        providerReserved: null,
-                                                                                        longForm: false,
-                                                                                        testChar);
 
-                    if (example == null) { continue; }
+                    if (testChar == '{')
+                    {
+                        int z = 10;
+                    }
 
-                    if (++count == 20)
+                    string example;
+
+                    try
+                    { 
+                        example = IdentifiableSecrets.GenerateCommonAnnotatedTestKey(IdentifiableSecrets.VersionTwoChecksumSeed,
+                                                                                     "TEST",
+                                                                                     customerManagedKey: true,
+                                                                                     platformReserved: null,
+                                                                                     providerReserved: null,
+                                                                                     longForm,
+                                                                                     testChar);
+                    }
+                    catch (InvalidOperationException)
+                    {
+                        example = null;
+                    }
+
+                    if (++count > 26)
                     {
                         break;
                     }
+
+                    if (example == null) { continue; }
 
                     yield return example;
                 }
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
@@ -19,7 +19,6 @@ namespace Microsoft.Security.Utilities
 
         public override IEnumerable<string> GenerateTruePositiveExamples()
         {
-            int count = 0;
             int attempts = 0;
 
             foreach (bool longForm in new[] { true, false })
@@ -30,7 +29,7 @@ namespace Microsoft.Security.Utilities
 
                     if (testChar == '{')
                     {
-                        int z = 10;
+                        break;
                     }
 
                     string example;
@@ -48,11 +47,6 @@ namespace Microsoft.Security.Utilities
                     catch (InvalidOperationException)
                     {
                         example = null;
-                    }
-
-                    if (++count > 26)
-                    {
-                        break;
                     }
 
                     if (example == null) { continue; }
--- a/src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
+++ b/src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
@@ -58,6 +58,7 @@ public static class WellKnownRegexPatterns
 
     public static IEnumerable<RegexPattern> HighConfidenceMicrosoftSecurityModels { get; } = new RegexPattern[]
     {
+        new CommonAnnotatedSecurityKey(),
         new AadClientAppIdentifiableCredentials(),
         new AzureFunctionIdentifiableKey(),
         new AzureSearchIdentifiableQueryKey(),
--- a/src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
+++ b/src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
@@ -58,7 +58,7 @@ public static class WellKnownRegexPatterns
 
     public static IEnumerable<RegexPattern> HighConfidenceMicrosoftSecurityModels { get; } = new RegexPattern[]
     {
-        new CommonAnnotatedSecurityKey(),
+        //new CommonAnnotatedSecurityKey(),
         new AadClientAppIdentifiableCredentials(),
         new AzureFunctionIdentifiableKey(),
         new AzureSearchIdentifiableQueryKey(),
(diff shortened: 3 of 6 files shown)


src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
```diff
@@ -22,26 +22,41 @@ public override IEnumerable<string> GenerateTruePositiveExamples()
             int count = 0;
             int attempts = 0;
 
-            while(true)
+            foreach (bool longForm in new[] { true, false })
             {
-                foreach (bool longForm in new[] { true, false })
+                while (true)
```
michaelcfanning, line 26: >while [](http://example.com/codeflow?start=16&length=5)

Good job moving this into the foreach loop! wow

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
```diff
@@ -22,26 +22,41 @@ public override IEnumerable<string> GenerateTruePositiveExamples()
...
-                                                                                        customerManagedKey: true,
-                                                                                        platformReserved: null,
-                                                                                        providerReserved: null,
-                                                                                        longForm: false,
-                                                                                        testChar);
 
-                    if (example == null) { continue; }
+                    if (testChar == '{')
+                    {
+                        int z = 10;
```
michaelcfanning: >int [](http://example.com/codeflow?start=24&length=3)

why don't we just use this condition to break? i.e., remove the count. 

basically, we just scan a - z and halt once we hit the brace.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
```diff
@@ -22,26 +22,41 @@ public override IEnumerable<string> GenerateTruePositiveExamples()
...
+                                                                                     longForm,
+                                                                                     testChar);
+                    }
+                    catch (InvalidOperationException)
                     {
-                        break;
+                        example = null;
                     }
 
+                    if (++count > 26)
```
michaelcfanning: >26 [](http://example.com/codeflow?start=34&length=2)

this kind of magic number is really no good in code. see my other suggestion, just break once we've examined all the chars that are legal for this scenario, lowercase alpha

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
suvamM: Fixed.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
suvamM: Fixed.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
```diff
@@ -19,28 +19,37 @@ public CommonAnnotatedSecurityKey()
...
+                    { 
+                        example = IdentifiableSecrets.GenerateCommonAnnotatedTestKey(IdentifiableSecrets.VersionTwoChecksumSeed,
+                                                                                     "TEST",
+                                                                                     customerManagedKey: true,
+                                                                                     platformReserved: null,
+                                                                                     providerReserved: null,
+                                                                                     longForm,
+                                                                                     testChar);
+                    }
+                    catch (InvalidOperationException)
```
michaelcfanning, line 47: >InvalidOperationException [](http://example.com/codeflow?start=27&length=25)

This seems wrong. The real problem here is that someone specified a test char that does not comprise a base62 character. We should change the API to perform this check and raise `ArgumentException`. If you prefer not to be slowed down by this suggestion, feel free to check in now and defer that change.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -58,6 +58,7 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
 
     public static IEnumerable<RegexPattern> HighConfidenceMicrosoftSecurityModels { get; } = new RegexPattern[]
     {
+        new CommonAnnotatedSecurityKey(),
```
michaelcfanning, line 64: >CommonAnnotatedSecurityKey [](http://example.com/codeflow?start=12&length=26)

I am not approving your change because it did not come with any new tests.

We clearly have a test gap, yes? :) because this was missing. please think about how you might close that test gap.

michaelcfanning (changes requested):
:clock1:

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
suvamM: I agree this is a better approach, since it will avoid a bunch of processing on an illegal character before the argument is thrown. I'll track this and make follow-on change.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
suvamM: @michaelcfanning Pushing back on this. We need this tactical fix. I'll track this and figure out how to close the larger test gap.

michaelcfanning: I've got it, will add a test for you.

michaelcfanning: Done. Checked in a failing test for you to examine and resolve. Please note the extremely worrisome lurking bugs this test uncovered. :)

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -54,10 +54,12 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
         new Unclassified32ByteBase64String(),
         new Unclassified64ByteBase64String(),
         new Unclassified16ByteHexadecimalString(),
+        // new UrlCredentials()
```
shaopeng-gh, line 59: >// new UrlCredentials() [](http://example.com/codeflow?start=8&length=23)

remove?

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -78,6 +80,7 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
         new AzureApimIdentifiableRepositoryKey(),
         new AzureCacheForRedisIdentifiableKey(),
         new AzureContainerRegistryIdentifiableKey(),
+        // new AzureContainerRegistryLegacyKey(),
```
shaopeng-gh, line 87: >// new AzureContainerRegistryLegacyKey(), [](http://example.com/codeflow?start=8&length=41)

remove?

shaopeng-gh (dismissed):
:shipit:

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
suvamM: This needs to be added in. It's commented because the rule is broken.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
suvamM: This needs to be added in. It's commented since the rule is broken.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -54,10 +54,12 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
         new Unclassified32ByteBase64String(),
         new Unclassified64ByteBase64String(),
         new Unclassified16ByteHexadecimalString(),
+        // new UrlCredentials()
     };
 
     public static IEnumerable<RegexPattern> HighConfidenceMicrosoftSecurityModels { get; } = new RegexPattern[]
     {
+        new CommonAnnotatedSecurityKey(),
```
yongyan-gh, line 64: >CommonAnnotatedSecurityKey [](http://example.com/codeflow?start=12&length=26)

nit: order by alphabet

yongyan-gh: do these keys needs to be sync with what SPMI has?
I see some of them are out of sync.
e.g. AzureEventGrid has both v1 and v2 in SPMI

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,9 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
 
+# UNRELEASED
```
michaelcfanning: >UNRELEASED [](http://example.com/codeflow?start=2&length=10)

Add this rule to well known patterns

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,9 @@
 - FPS => False positive reduction in static analysis.
 - FNS => Flase negative reduction in static analysis.
```
michaelcfanning: >Flase [](http://example.com/codeflow?start=9&length=5)

spell this correctly

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
michaelcfanning, line 59: >UrlCredentials [](http://example.com/codeflow?start=15&length=14)

add a comment please

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -54,10 +54,12 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
         new Unclassified32ByteBase64String(),
         new Unclassified64ByteBase64String(),
         new Unclassified16ByteHexadecimalString(),
+        // new UrlCredentials()
     };
 
     public static IEnumerable<RegexPattern> HighConfidenceMicrosoftSecurityModels { get; } = new RegexPattern[]
```
michaelcfanning, line 62: leave a human readable comment

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
suvamM: Added comment to GH issue

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
suvamM: I left this open---it deserves a follow on PR as it's an API change

PR #47, opened 2024-06-14
Adding support for .NET 451 for backwards compatibility

suvamM:
Some consumers of `security-utilities` are on older versions of .Net. This PR adds support for .net451 in `Microsoft.Security.Utilities.Core` to improve backward compatibility.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -226,7 +226,8 @@ public static class IdentifiableSecrets
             if (testChar == null)
             {
                 using var generator = RandomNumberGenerator.Create();
-                generator.GetBytes(keyBytes, 0, (int)keyLengthInBytes);
+
+                generator.GetBytes(keyBytes);
 
                 key = keyBytes.ToBase62();
                 
@@ -391,7 +392,7 @@ public static class IdentifiableSecrets
         byte[] randomBytes = new byte[(int)keyLengthInBytes];
 
         using var generator = RandomNumberGenerator.Create();
-        generator.GetBytes(randomBytes, 0, (int)keyLengthInBytes);
+        generator.GetBytes(randomBytes);
 
         string secret = GenerateBase64KeyHelper(checksumSeed,
                                                 keyLengthInBytes,
@@ -422,7 +423,7 @@ public static class IdentifiableSecrets
         byte[] randomBytes = new byte[(int)keyLengthInBytes];
 
         using var generator = RandomNumberGenerator.Create();
-        generator.GetBytes(randomBytes, 0, (int)keyLengthInBytes);
+        generator.GetBytes(randomBytes);
 
         return GenerateBase64KeyHelper(checksumSeed,
                                        keyLengthInBytes,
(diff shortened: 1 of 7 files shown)


src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -226,7 +226,8 @@ public static ulong ComputeHisV1ChecksumSeed(string versionedKeyKind)
             if (testChar == null)
             {
                 using var generator = RandomNumberGenerator.Create();
-                generator.GetBytes(keyBytes, 0, (int)keyLengthInBytes);
+
+                generator.GetBytes(keyBytes);
```
suvamM, line 230: In all the usages of `GetBytes`, the byte array length is precisely equal to the number of bytes required. So the 3 argument `GetBytes` API is not strictly required.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_102_AdoPat.cs
```diff
@@ -73,7 +73,7 @@ private static byte[] ConvertFromBase32(string inputString)
 
             if (string.IsNullOrEmpty(inputString))
             {
-                return Array.Empty<byte>();
+                return new byte[0];
```
michaelcfanning: >new [](http://example.com/codeflow?start=23&length=3)

you should create a static readonly instance for this. why allocate each time?

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_106_AzureStorageAccountLegacyCredentials.cs
```diff
@@ -72,7 +72,7 @@ private static byte[] ConvertFromBase32(string inputString)
 
             if (string.IsNullOrEmpty(inputString))
             {
-                return Array.Empty<byte>();
+                return new byte[0];
```
michaelcfanning: >new [](http://example.com/codeflow?start=23&length=3)

DRY out your code `don't repeat yourself`. declare a static readonly empty byte array in a logical place and utilize it from the two rule locations you have found.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_102_AdoPat.cs
suvamM: Excellent insight. Making a change.

src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <IsPackable>true</IsPackable>
-    <TargetFrameworks>net462;netstandard2.0;net6.0;net8.0</TargetFrameworks>
```
shaopeng-gh, line 5: >net462 [](http://example.com/codeflow?start=22&length=6)

in the ps1 there is
`$Configuration\net462\SecurityUtilitiesApiUtilizationExample.exe`
do we need to keep and add instead of replace?

suvamM: @shaopeng-gh Yea that just adds validation for the compatibility.

shaopeng-gh (approved):
:shipit:

PR #50, opened 2024-06-17
Base62 threading issue

michaelcfanning:
- BUG: Resolve `System.ArgumentOutOfRangeException: Index was out of range` and `System.FormatException: The input is not a valid Base-46 string` errors when calling `IdentifiableSecrets.GenerateCommonAnnotatedTestKey(ulong, string, bool, byte[], byte[], bool, char?)`. These exceptions originated in multithreading issues in `Base62.EncodingExtensions.ToBase62(this string)`.
- PRF: Eliminate instantiation of `RandomNumberGenerator` object on every key allocation.

### DIFF
--- a/src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
+++ b/src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
@@ -326,7 +326,7 @@ namespace Microsoft.Security.Utilities
         [TestMethod]
         public void IdentifiableSecrets_PlatformAnnotatedSecurityKeys()
         {
-            int iterations = 100;
+            int iterations = 10;
             ulong keysGenerated = 0;
             const string alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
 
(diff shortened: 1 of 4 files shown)


src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -325,76 +326,88 @@ public void IdentifiableSecrets_DerivedSymmetricKeys()
...
+
                     for (byte k = 0; k < iterations; k++)
                     {
                         string signature = Guid.NewGuid().ToString("N").Substring(0, 4);
+                        signature = $"{alphabet[(int)keysGenerated % alphabet.Length]}{signature.Substring(1)}";
+                        signatures.Add(signature);
+                    }
 
-                        signature = $"{alphabet[(int)keysGenerated % alphabet.Length]}{ signature.Substring(1)}";
+                    Parallel.ForEach(signatures, signature =>
```
michaelcfanning, line 346: >ForEach [](http://example.com/codeflow?start=29&length=7)

This threading change predictably flushed out our threading issue.

suvamM (approved):

rwoll (approved):

PR #54, opened 2024-06-18
preparing versions for 1.5.0

suvamM:


version.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.4.22",
```
shaopeng-gh, line 3: Was this a bug that it was different ?

shaopeng-gh: Is there other "1.4.22" in source code that need to be updated:
image

suvamM: @shaopeng-gh yes. we missed updating the version.

shaopeng-gh: The test show all passed, was the test not effective, or actually testing something else not related
image

suvamM: @shaopeng-gh This test does not seem to execute. But it's a valid concern and I will follow up on what the deal here is.

shaopeng-gh (approved):
:shipit:

PR #55, opened 2024-06-18
Eliminate a rule from the WellKnownRegexPatterns list

suvamM:
This rule was already eliminated, but persisted in the `WellKnownRegexPatterns` list.


shaopeng-gh (approved):
:shipit:

PR #56, opened 2024-06-18
updating automated regex pattern json

suvamM:


GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
```diff
@@ -1,11 +1,4 @@
 [
-  {
```
shaopeng-gh, line 2: Is this file auto gen?

suvamM: Yes. But it's also checked in.

shaopeng-gh (approved):

PR #59, opened 2024-06-28
Perf Benchmark

suvamM:
This PR adds a test which does the following:

1. Generate 1M random test keys using the common annotated security key generator.
2. For each test key, run the high performance identifiable scanning---timing the scan and result extraction.
3. For each result, perform checksum validation---timing the validation.

Then print the mean scan and validation times. This allows us to measure the overhead of performing checksum validation.

### DIFF
--- a/src/security_utilities_rust/src/end_to_end_tests.rs
+++ b/src/security_utilities_rust/src/end_to_end_tests.rs
@@ -34,3 +34,65 @@ fn generate_and_detect_common_annotated_key_test() {
 
     assert_eq!(validation_result, true, "checksum validation of generated input passes");
 }
+
+#[test]
+fn identifiable_scanning_and_validation_perf_benchmark() {
+    use std::time::{Duration, Instant};
+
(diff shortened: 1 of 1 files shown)


rwoll (commented):
This looks like a bespoke benchmark instead of an idiomatic Rust Benchmark using built-in cargo bench: https://doc.rust-lang.org/1.4.0/book/benchmark-tests.html

Please update to use cargo's benchmarking tool or abandon the PR.

suvamM:
> This looks like a bespoke benchmark instead of an idiomatic Rust Benchmark using built-in cargo bench: https://doc.rust-lang.org/1.4.0/book/benchmark-tests.html
> 
> Please update to use cargo's benchmarking tool or abandon the PR.

@rwoll This does not look like a stable feature of the Rust toolchain currently. From the link you shared:

```
Note the test feature gate, which enables this unstable feature.
```
This benchmark is important to have checked in. So I'll track the `bench` verb (thanks for the pointer!), and will move it when it becomes stable.

rwoll (approved):

PR #61, opened 2024-07-03
Adding an initial secret masking capability

suvamM:
This PR adds an initial secret masking capability to Rust. For a given input, the masker leverages the high performance scanner to compute finds, optionally performs checksum validation, and then redacts finds. The redaction token can be specified by the user, otherwise it defaults to `SEC101/200 : c3id`, where `c3id` is the cross-company correlating id for the find.

The PR also adds a test to highlight the usage of the API, and benchmarks the performance with and without checksum validation. On my machine, the performance was:

Mean time for masking with checksum validation: 97.798µs
Mean time for masking without checksum validation: 80.614µs

_Edit_
I implemented an "in-place" redaction, based on PR feedback from @beaubelgrave . The resulting performance is:
Mean time for masking with checksum validation: 66.662µs
Mean time for masking without checksum validation: 55.299µs

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
@@ -102,11 +102,12 @@ impl ScanMatch {
     }
 }
 
+#[derive(Clone)]
 pub struct PossibleScanMatch {
     name: &'static str,
     def_index: u32,
-    start: u64,
-    len: usize,
+    pub start: u64,
+    pub len: usize,
     utf8: bool,
     validator: Rc<dyn Fn(&[u8]) -> usize>,
 }
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
@@ -106,8 +106,8 @@ impl ScanMatch {
 pub struct PossibleScanMatch {
     name: &'static str,
     def_index: u32,
-    pub start: u64,
-    pub len: usize,
+    start: u64,
+    len: usize,
     utf8: bool,
     validator: Rc<dyn Fn(&[u8]) -> usize>,
 }
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
@@ -827,9 +827,11 @@ pub struct SecretMasker {
 }
 
 impl SecretMasker {
-    pub fn mask_secrets(&mut self, input: &str, default_redaction_token: Option<&str>, validate_checksum: bool) -> String {
+    pub fn mask_secrets(&mut self, input: &mut String, default_redaction_token: Option<&str>, validate_checksum: bool) -> Option<String> {
+        let default_return = String::new();
+
         if input.is_empty() {
-            return String::new();
+            return Some(default_return);
         }
 
         self.scan.reset();
@@ -838,18 +840,19 @@ impl SecretMasker {
         self.scan.parse_bytes(input_as_bytes);
 
         let detections = self.scan.possible_matches();
-        let mut detections = detections.clone();
 
         // Short-circuit if nothing to replace.
         if detections.is_empty() {
-            return input.to_string();
+            return None;
         }
+
+        let mut detections = detections.clone();
 
         // Merge positions into ranges of characters to replace.
         let mut current_detections = Vec::new();
         let mut current_detection : Option<Detection> = None;
 
-        detections.sort_unstable_by_key(|item| item.start);
+        detections.sort_unstable_by_key(|item| item.start());
 
         for detection in detections.iter() {
             let scan_match = detection.matches_bytes(input_as_bytes, true).unwrap();
@@ -888,9 +891,8 @@ impl SecretMasker {
                     redaction_token: redaction_token.to_string()
                 });
                 current_detections.push(current_detection.clone().unwrap());
-            }
-            else {
-                if detection.start <= current_detection.clone().unwrap().end {
+            } else {
+                if detection.start() <= current_detection.clone().unwrap().end {
                     // Overlapping case or contiguous case.
                     let current = current_detection.as_mut().unwrap();
                     current.length = (std::cmp::max(current.end,
@@ -908,19 +910,14 @@ impl SecretMasker {
             }
         }
 
-        let mut result = String::new();
-        let mut start_index = 0;
+        let mut index_adjustment: usize = 0;
         for detection in current_detections {
-            result.push_str(&input[start_index as usize..detection.start as usize]);
-            result.push_str(&detection.redaction_token);
-            start_index = detection.start + detection.length as u64;
+            let start_index = detection.start as usize - index_adjustment;
+            input.replace_range(start_index..start_index + detection.length, &detection.redaction_token);
+            index_adjustment += detection.length - detection.redaction_token.len();
         }
 
-        if start_index < input.len() as u64 {
-            result.push_str(&input[start_index as usize..]);
-        }
-
-        result
+        Some(default_return)
     }
 }
 
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
@@ -11,12 +11,15 @@ use chrono::Datelike;
 use core::panic;
 use lazy_static::lazy_static;
 use std::{mem};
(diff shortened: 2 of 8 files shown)


shaopeng-gh:
Glad that the test is fixed, 
not familiar with Rust, but LGTM

shaopeng-gh (dismissed):
:shipit:

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -102,11 +102,12 @@ impl ScanMatch {
...
 }
 
+#[derive(Clone)]
 pub struct PossibleScanMatch {
     name: &'static str,
     def_index: u32,
-    start: u64,
-    len: usize,
+    pub start: u64,
+    pub len: usize,
```
beaubelgrave: It's preferable to have a start() and len() pub method vs exposing pub fields directly. Rust idiomatic approach is that, but sometimes I cut that corner.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -811,3 +814,114 @@ fn index_of(input_string: String, search_char: char) -> i32
...
+    length: usize,
+    redaction_token: String
+}
+
+pub struct SecretMasker {
+    pub scan: identifiable_scans::Scan
+}
+
+impl SecretMasker {
+    pub fn mask_secrets(&mut self, input: &str, default_redaction_token: Option<&str>, validate_checksum: bool) -> String {
```
beaubelgrave: It's not immediate, but I would like to see a way to mask_secrets on files that are binary and not a text string. We have large binary files that could use this masking as well.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -811,3 +814,114 @@ fn index_of(input_string: String, search_char: char) -> i32
...
+#[derive(Clone)]
+struct Detection {
+    start: u64,
+    end: u64,
+    length: usize,
+    redaction_token: String
+}
+
+pub struct SecretMasker {
+    pub scan: identifiable_scans::Scan
```
beaubelgrave, line 826: Does this need to be pub?

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -811,3 +814,114 @@ fn index_of(input_string: String, search_char: char) -> i32
...
+                        end: detection.start() + detection.len() as u64,
+                        length: detection.len(),
+                        redaction_token: redaction_token.to_string()
+                    });
+                    current_detections.push(current_detection.clone().unwrap());
+                }
+            }
+        }
+
+        let mut result = String::new();
```
beaubelgrave: It would be nice to have an in-place masker instead of text copying. This will work fine for small lines of text, but not large lines of text without overheads.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -811,3 +814,114 @@ fn index_of(input_string: String, search_char: char) -> i32
...
+
+        let input_as_bytes = input.as_bytes();
+        self.scan.parse_bytes(input_as_bytes);
+
+        let detections = self.scan.possible_matches();
+        let mut detections = detections.clone();
+
+        // Short-circuit if nothing to replace.
+        if detections.is_empty() {
+            return input.to_string();
```
beaubelgrave: This causes an allocation and string copy. When there are no secrets, we want that to be fast, right?

beaubelgrave: I would suggest the function returning an Option<String> if you want to not have in-place masking.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -811,3 +814,114 @@ fn index_of(input_string: String, search_char: char) -> i32
...
+            return String::new();
+        }
+
+        self.scan.reset();
+
+        let input_as_bytes = input.as_bytes();
+        self.scan.parse_bytes(input_as_bytes);
+
+        let detections = self.scan.possible_matches();
+        let mut detections = detections.clone();
```
beaubelgrave: Probably don't need to clone if it's empty, which should increase the speed very slightly. Just calling it out.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -811,3 +814,114 @@ fn index_of(input_string: String, search_char: char) -> i32
...
+            if current_detection.is_none() {
+                current_detection = Some(Detection {
+                    start: detection.start(),
+                    end: detection.start() + detection.len() as u64,
+                    length: detection.len(),
+                    redaction_token: redaction_token.to_string()
+                });
+                current_detections.push(current_detection.clone().unwrap());
+            }
+            else {
```
beaubelgrave: else should be on the line above, IE: } else {

suvamM: Fixed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
suvamM: `start()` and `len()` was already there and I had missed it. Fixed: I now leverage these public methods instead.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
suvamM: Valid point. Fixed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
suvamM: Valid suggestion. Fixed.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
suvamM: @beaubelgrave Interesting. So the overhead scenario might be something like: `bunch of text` <1 secret> `another bunch of text`. In this case, with the current implementation, we would needlessly end up copying both the text bunches, even though we just needed a single replacement. I guess the C# implementation also suffers from this performance bottleneck. 
Let me take a stab at optimizing the Rust method based on your feedback.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
suvamM: @beaubelgrave Tracking via https://github.com/microsoft/security-utilities/issues/62

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
suvamM: OK @beaubelgrave I took a stab at implementing an "in-place" masker. With this, there was a 31% reduction in the masking time. Look forward to your thoughts on this.

@michaelcfanning for awareness. If this implementation is finalized, we may want to modify the C# implementation based along the Rust algorithm.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -811,3 +814,111 @@ fn index_of(input_string: String, search_char: char) -> i32
...
+                    });
+                    current_detections.push(current_detection.clone().unwrap());
+                }
+            }
+        }
+
+        let mut index_adjustment: usize = 0;
+        for detection in current_detections {
+            let start_index = detection.start as usize - index_adjustment;
+            input.replace_range(start_index..start_index + detection.length, &detection.redaction_token);
```
suvamM, line 914: @beaubelgrave I am piggybacking on `replace_range` to do the in-place replacement, followed by some math adjustment on the indices. Is this a good choice?

beaubelgrave: I think so.

Since you modify the input now, I wouldn't return an Option<String>. Maybe just return a bool indicating if the text was modified/changed?

If I were to read this function signature I would have thought the Option<String> would be returning something like the modified string, but I don't think that's the intention anymore.

shaopeng-gh (approved):

PR #64, opened 2024-07-12
Rust compiler E0716 error fix

suvamM:
This PR makes a fix for `E0716: a temporary value is being dropped while a borrow is still in active use`.


yongyan-gh (approved):

scottoneil-ms (commented):
LGTM

scottoneil-ms (approved):

PR #66, opened 2024-07-16
Rust: Fix benchmark performance test

beaubelgrave:
The current benchmark test for the Rust scanner omitted the scanner getting reset. This causes the scanner to accumulate possible matches and cause the Vec to resize. This leads to an inaccurate picture of performance of the scanner.

By properly resetting the scanner, the performance increases by ~20%.


michaelcfanning (approved):
:shipit:

PR #67, opened 2024-07-18
Optimize secret masking, c3id genertion, `marvin`, and checksum validation.

jdraaijer-microsoft:
This PR improves the `secret_masker_test`  with checksum benchmark by about 32%, from 1300 ns to 880 ns per iteration, and the without checksum by about 15% from 879ns to 734 ns per iteration, on my machine.

It applies the following changes:
* `marvin`
  * Made more idiomatic by relying on slices to provide us with lengths and offsets (~20% speedup). Updated existing impl to use that after validating slice constraints.
  * Applied `rustfmt`
  * Increased test coverage by adding offsets and non-full-length tests
* Cross company IDs
  * Decreased amount of work required by removing some unnecessary formatting calls and pre-allocating data for known-length strings.
* Secret masking
  * During validation, skip validating the correctness of the key: if it contains invalid bytes, the signature check will fail and we panic either way, so the error message is lost. If we do want this, we should add it as a "post" step, where we check the validity of the signature bytes in the case that validation fails and report it to the caller. It may be that I misunderstood the point of this check, so please do leave feedback in case it is necessary :)
  * Don't calculate the cross-company ID if we don't need it
  * Use the `last()` detection to track the current one instead of keeping separate state + cloning and unwrapping a bunch
* Scanning
  * API BREAKING (but functionality remains the same): `ScanMatch`: use `Cow<'a, str>` instead of `String` to avoid unnecessarily copying utf8 data to a new string.
* identifiable secrets
  * Make a bunch of `static`s `const`.
  * Remove checks that are expensive on the happy path and have minor perf improvement for the un-happy path, that are not explicitly reported.
  * Decrease amount of en- and decoding required for validating a key signature by re-using existing bytes.

This may be (far too) much for a single PR, but I have tried my best to split the changes into concise commits and am willing to make separate PRs for them if that is preferred.

### DIFF
--- a/src/security_utilities_rust/src/identifiable_secrets_tests.rs
+++ b/src/security_utilities_rust/src/identifiable_secrets_tests.rs
@@ -30,7 +30,13 @@ fn secret_masker_test() {
         scan:  microsoft_security_utilities_core::identifiable_scans::Scan::new(options)
     };
 
-    for _ in 0..1000 {
+    #[cfg(debug_assertions)]
+    let iterations = 1000;
+
+    #[cfg(not(debug_assertions))]
+    let iterations = 500_000;
+
+    for _ in 0..iterations {
         // generate a key
         let valid_key = microsoft_security_utilities_core::identifiable_secrets::
         generate_common_annotated_key(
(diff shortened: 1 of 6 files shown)


jdraaijer-microsoft:
I've updated the commit messages to be more descriptive, and make it obvious which commits touch which parts.

suvamM:
@jdraaijer-microsoft This PR has some good changes. But reviewing this is hard for me. Could you please scope the changes into smaller PRs?

jdraaijer-microsoft:
Closing in favor of smaller (future) PRs: #69, #70

PR #68, opened 2024-07-18
Rust/His: Allow for concurrent scans

beaubelgrave:
Today the existing Scan struct tracks the options to be used as well as the current state of the stream. This is useful when doing many streams of data on a per-core basis. It is not as useful if you are using many threads.

When many threads are being used, it's more useful to have a single immutable reference that can be passed small mutable states that can individually be reset or thrown away.

Add ScanState struct that solely keeps track of the state of a given stream of data.

Add concurrent_parse_bytes() and concurrent_parse_reader() functions that take a ScanState mutable reference while operating on the immutable Scan struct. This allows for many threads to share the same Scan struct.

Move the internal scanning state within the Scan struct to a RefCell ScanState. This allows us to keep our API the same and not break our FFI layers while allowing for immutable changes to the byte scan.

Move the various byte scan logic methods to immutable self references that take in a mutable ScanState reference.

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 use std::rc::Rc;
+use std::cell::RefCell;
 
 /* Indicates the char is part of a small mask */
 const MASK_SMALL: u8 = 1 << 0;
@@ -847,30 +848,47 @@ impl Default for ScanOptions {
     }
 }
 
+#[derive(Default)]
+pub struct ScanState {
+    pub accum: u64,
+    pub index: u64,
+    pub must_scan: bool,
+    pub checks: Vec<PossibleScanMatch>,
+}
+
+impl ScanState {
+    pub fn reset(&mut self) {
+        self.accum = 0;
+        self.index = 0;
+        self.must_scan = false;
+        self.checks.clear();
+    }
+
+    pub fn has_possible_matches(&self) -> bool { !self.possible_matches().is_empty() }
+
+    pub fn possible_matches(&self) -> &Vec<PossibleScanMatch> { &self.checks }
+}
+
 pub struct Scan {
     options: ScanOptions,
-    accum: u64,
-    index: u64,
+    stream_state: RefCell<ScanState>,
     checks: Vec<PossibleScanMatch>,
     utf8_lanes: [Vec<ScanDefinition>; 32],
     utf16_lanes: [Vec<ScanDefinition>; 32],
     sig_char_chunks: Vec<[u8; 16]>,
     char_map: [u8; 256],
-    must_scan: bool,
 }
 
 impl Scan {
     pub fn new(options: ScanOptions) -> Self {
         let mut scan = Self {
             options,
-            accum: 0,
-            index: 0,
+            stream_state: RefCell::new(ScanState::default()),
             checks: Vec::new(),
             utf8_lanes: Default::default(),
             utf16_lanes: Default::default(),
             sig_char_chunks: Vec::new(),
             char_map: [0; 256],
-            must_scan: false,
         };
 
         scan.init();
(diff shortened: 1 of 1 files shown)


src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -847,30 +848,47 @@ impl Default for ScanOptions {
...
+    pub fn has_possible_matches(&self) -> bool { !self.possible_matches().is_empty() }
+
+    pub fn possible_matches(&self) -> &Vec<PossibleScanMatch> { &self.checks }
+}
+
 pub struct Scan {
     options: ScanOptions,
-    accum: u64,
-    index: u64,
+    stream_state: RefCell<ScanState>,
```
michaelcfanning, line 874: >stream_state [](http://example.com/codeflow?start=4&length=12)

it's not a big deal but maybe `scan_states` is a better name here? because this object contains non-stream related data (like the accumulated checks)

beaubelgrave: Maybe, that ScanState is only used for the "stream" interfaces.

I'm not sure what a good name is for them, I'm happy with scan_state, or internal_state.
It's not used at all when concurrent_* APIs are used by external callers.

michaelcfanning (approved):
:shipit:

michaelcfanning:
We probably should start investing in a change log, such as we have in `ReleaseHistory.md` for the C# code. I will look at the C# implications for your change tomorrow, but we can merge this if you like.

jdraaijer-microsoft:
I have a stylistic/bikeshedding question, as I'm wondering (but feel free to disregard):

I think having both the concurrent and non-concurrent functions on `Scan` creates some API disparity, as one the non-concurrent ones update some internal state on `Scan`, while the concurrent ones do not. It's obvious from the function signatures, but not super necessary. Is there a specific reason why you've decided against creating an additional "scan configuration" struct (besides the fact that it is less code to change) that would be responsible for the concurrent functions?

Additionally, storing the existing `PossibleMatch`es outside of the `ScanState` struct could let you avoid the `RefCell` alltogether, and give a caller more freedom w.r.t. what they want to do with the existing matches (without having to `reset` the entire state).

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
michaelcfanning: I see. let's leave it as `stream_state`.

beaubelgrave:
> I have a stylistic/bikeshedding question, as I'm wondering (but feel free to disregard):
>

Sure.

> I think having both the concurrent and non-concurrent functions on `Scan` creates some API disparity, as one the non-concurrent ones update some internal state on `Scan`, while the concurrent ones do not. It's obvious from the function signatures, but not super necessary. Is there a specific reason why you've decided against creating an additional "scan configuration" struct (besides the fact that it is less code to change) that would be responsible for the concurrent functions?
> 

The main reason is we already have a release, and that includes an FFI that assumes certain things. It's possible to move everything to the concurrent_* versions. The safest route, IMHO, was to make two different versions of the API, keep the FFI using the existing pattern, and let newer code choose if they want that level of management or not.

This would be like how the Read trait in Rust kind of evolved. You have the basic read(), then you have read_exact(), read_to_end() etc, that leverage that single read() base call. And like the Seek trait, where the objects typically store their own offset, but sometimes you can read directly without the Seek part (IE: Memory based reader).

> Additionally, storing the existing `PossibleMatch`es outside of the `ScanState` struct could let you avoid the `RefCell` alltogether, and give a caller more freedom w.r.t. what they want to do with the existing matches (without having to `reset` the entire state).

RefCell overhead is minimal and only when using the stream processor mode, when concurrent is used there is no overhead of the RefCell (other than its mere existence in the Scan struct). Many scenarios I know about require to know all the possible matches for redactions or auditing. I've not come across anything where you'd really want to split these out. If we do come across that need, a simple take_possible_matches() can be added to the ScanState.

Hope this helps clarify at least where I'm coming from regarding this.

PR #69, opened 2024-07-22
Rust/c3id: Decrease allocations & formatting

jdraaijer-microsoft:
Reducing the amount of required allocations improves performance by about 20% for few-block computations (0 to 88 characters), and by about 6% for many block (1024 characters) due to reduced (re-)allocation overhead.

[Benchmark](https://github.com/datdenkikniet/security-utilities/blob/bench/src/security_utilities_rust/benches/cross_company_correlating_id.rs)


jdraaijer-microsoft:
Closing in favor of https://github.com/microsoft/security-utilities/pull/74

PR #70, opened 2024-07-22
Rust/Marvin: add slice-only functions

jdraaijer-microsoft:
Instead of re-using the offset/length arguments, we can simply construct a slice and operate on that, making the code a lot more concise and about 33% faster.

Also increased test coverage to test non-default offsets and lengths for correct output.

The first commit adds only the slice methods & equality tests, the second one modifies the tests more extensively.

[Benchmark](https://github.com/datdenkikniet/security-utilities/blob/bench/src/security_utilities_rust/benches/marvin.rs)


jdraaijer-microsoft:
Closing in favor of #73

PR #72, opened 2024-07-23
Rust/His: Introduce ScanEngine struct

beaubelgrave:
Previously the Scan struct would track it's own state as well as allow external state operations. To do this though, it required a RefCell instance for the non-external state. This prevented the Scan struct from being fully used in scoped thread contexts without a Mutex or Arc.

Introduce the ScanEngine struct which contains all the same code as the Scan struct (rename) as before except for the internal state tracking.

Re-write the Scan struct to simply host the ScanEngine and a ScanState. This ensures the logic is the same as before, however, now we can completely drop the RefCell and make it much clearer for callers what we expect them to do.

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -13,6 +13,7 @@
 
 # UNRELEASED
 - DEP: `System.Text.Json` updated to `v8.0.4` to resolve Depandabot alert.
+- BRK: `Scan` struct now comprises a `ScanEngine` and a `ScanState` instance. Scan information, such as `checks`, must be accessed via the `ScanState` field of the `Scan` struct. 
 
 # 1.5.2 - 07/05/2024
 - NEW: Added an initial secret redaction capability to the Rust package.
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 use std::rc::Rc;
-use std::cell::RefCell;
 
 /* Indicates the char is part of a small mask */
 const MASK_SMALL: u8 = 1 << 0;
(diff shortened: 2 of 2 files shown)


src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -1117,50 +1103,52 @@ impl Scan {
...
+    engine: ScanEngine,
+    state: ScanState,
+}
+
+impl Scan {
+    pub fn new(options: ScanOptions) -> Self {
+        Self {
+            engine: ScanEngine::new(options),
+            state: ScanState::default(),
+        }
```
suvamM, line 1124: @beaubelgrave This is a breaking change for existing customers right? Now users need to access the checks via the state. I will add a release note entry in this PR to highlight this breaking change.

suvamM, line 1124: @beaubelgrave check

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -1117,50 +1103,52 @@ impl Scan {
...
+pub struct Scan {
+    engine: ScanEngine,
+    state: ScanState,
+}
+
+impl Scan {
+    pub fn new(options: ScanOptions) -> Self {
+        Self {
+            engine: ScanEngine::new(options),
+            state: ScanState::default(),
```
suvamM, line 1123: @beaubelgrave This is a breaking change for existing users right? Now users need to access checks via the scan state object.
If my understanding is correct, then we need to add an entry in the release notes highlighting this. I can take care of the release note entry.

suvamM (dismissed):

docs/ReleaseHistory.md
```diff
@@ -13,6 +13,7 @@
 
 # UNRELEASED
 - DEP: `System.Text.Json` updated to `v8.0.4` to resolve Depandabot alert.
+- BRK: `Scan` struct now comprises a `ScanEngine` and a `ScanState` instance. Scan information, such as `checks`, must be accessed via the `ScanState` field of the `Scan` struct.
```
suvamM, line 16: @beaubelgrave This is the entry I added. Just a highlight about the breaking change. I'd be happy to update it if the entry is insufficient.

suvamM (approved):

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
jdraaijer-microsoft: AAIU this is not breaking: `checks` was only ever accessible externally through `{has_}possible_matches`, never directly. The API for `Scan` has not changed in that regard, only internally as some tests were accessing this field directly.

beaubelgrave: @jdraaijer-microsoft is correct, this isn't breaking. However, it's nice to call out the ability to use this for scans while owning the state for concurrent designs.

docs/ReleaseHistory.md
beaubelgrave: I don't think we need to mark it as BRK, but it's useful for people to know they can use the ScanEngine and ScanState to have truly concurrent designs without any locking, etc.

michaelcfanning: Given other comments here, I think I'd agree, this sounds like a `NEW`. A `BRK` designation is required for an update that may literally provoke a compilation error for a user (who is accessing public methods/data that has disappeared/been renamed, etc.) or if some runtime behavior has been changed in a fundamentally breaking way.

If the generally accessibly contract of an API hasn't changed and we've simply added the magic of permitting concurrency, this would be a new feature.

PR #73, opened 2024-07-23
Rust/Marvin: add slice-only functions

jdraaijer-microsoft:
Instead of re-using the offset/length arguments, we can simply construct a slice and operate on that, making the code a lot more concise and 20% faster.

Also increased test coverage to test non-default offsets and lengths for correct output.

The first commit adds only the slice methods & equality tests, the second one modifies the tests more extensively.

[Benchmark](https://github.com/microsoft/security-utilities/blob/users/jdraaijer/bench/src/security_utilities_rust/benches/marvin.rs)

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/marvin.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/marvin.rs
@@ -2,22 +2,17 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 /// This is a Rust implementation of the Marvin32 checksum algorithm, the definitive native code for which is
-/// at https://github.com/microsoft/SymCrypt/blob/master/lib/marvin32.c. 
-use std::mem;
+/// at https://github.com/microsoft/SymCrypt/blob/master/lib/marvin32.c.
 
 /// Convenience method to compute a Marvin hash and collapse it into a 32-bit hash.
-pub fn compute_hash32(data: &[u8], seed: u64, offset: i32, length: i32) -> i32
-{
-    let hash64: i64 = compute_hash(data, seed, offset, length);
+pub fn compute_hash32(data: &[u8], seed: u64, offset: i32, length: i32) -> i32 {
+    let hash64 = compute_hash(data, seed, offset, length);
     return ((hash64 >> 32) as i32) ^ (hash64 as i32);
 }
 
-/// Computes a 64-bit hash using the Marvin algorithm.
-pub fn compute_hash(data: &[u8], seed: u64, offset: i32, mut length: i32) -> i64
-{
-    // Marvin by design can produce a checksum for empty input buffers, which
-    // is why it's ok for the offset to point just past the end of the buffer
-    // or for the input buffer to be empty;
+/// Computes a 64-bit hash using the Marvin algorithm from the given slice,
+/// using the provided length and offset to determine the data to hash.
+pub fn compute_hash(data: &[u8], seed: u64, offset: i32, length: i32) -> i64 {
     if offset < 0 || offset > data.len() as i32
     {
         panic!("Offset '{}' is out of range", offset);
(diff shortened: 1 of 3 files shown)


src/security_utilities_rust/src/microsoft_security_utilities_core/marvin.rs
```diff
@@ -2,23 +2,24 @@
...
+/// Convenience method to compute a Marvin hash from a slice and collapse it into a 32-bit hash.
+pub fn compute_hash32_slice(data: &[u8], seed: u64) -> i32 {
+    let hash64 = compute_hash_slice(data, seed);
+    return ((hash64 >> 32) as i32) ^ (hash64 as i32);
+}
+
+/// Computes a 64-bit hash using the Marvin algorithm from the given slice,
+/// using the provided length and offset to determine the data to hash.
+pub fn compute_hash(data: &[u8], seed: u64, offset: i32, length: i32) -> i64 {
+    if offset > data.len() as i32
```
suvamM: >offset > data.len() as i32 [](http://example.com/codeflow?start=7&length=26)

The `offset < 0` was removed: this is a public function. What prevents someone from sending a negative offset?

suvamM (dismissed):

src/security_utilities_rust/src/microsoft_security_utilities_core/marvin.rs
jdraaijer-microsoft: Excellent catch, I missed that. Fixed

src/security_utilities_rust/src/microsoft_security_utilities_core/marvin.rs
```diff
@@ -2,22 +2,17 @@
...
 /// This is a Rust implementation of the Marvin32 checksum algorithm, the definitive native code for which is
-/// at https://github.com/microsoft/SymCrypt/blob/master/lib/marvin32.c. 
-use std::mem;
+/// at https://github.com/microsoft/SymCrypt/blob/master/lib/marvin32.c.
 
 /// Convenience method to compute a Marvin hash and collapse it into a 32-bit hash.
-pub fn compute_hash32(data: &[u8], seed: u64, offset: i32, length: i32) -> i32
-{
-    let hash64: i64 = compute_hash(data, seed, offset, length);
+pub fn compute_hash32(data: &[u8], seed: u64, offset: i32, length: i32) -> i32 {
```
michaelcfanning, line 8: >compute_hash32 [](http://example.com/codeflow?start=7&length=14)

Johannes, thanks for all these contributions. Can we add some release notes for various performance improvements? How are we benchmarking these changes? For the C# code, we have a good repeatable mechanism for verifying perf gains, what's our Rust story.

jdraaijer-microsoft: I will add some release notes later today :)

For the performance measurements, I'm using [`Criterion`][1]. I have written a few small benchmarks over on the [`users/jdraaijer/bench`][0] branch. It saves results and automagically compares the results from the previous run. Is this something that we may want to have a PR for so it can be merged and available to everyone?

[0]: https://github.com/microsoft/security-utilities/tree/users/jdraaijer/bench/src/security_utilities_rust
[1]: https://docs.rs/criterion/latest/criterion/

suvamM: @jdraaijer-microsoft Yes, it would be nice to have the benchmarks persisted, and documentation outlining how to repro. This can be done in a follow on PR. This PR definitely needs a release note.

jdraaijer-microsoft: I've added a release note for the changes in this PR, and opened https://github.com/microsoft/security-utilities/pull/79 for the benchmarks.

suvamM (approved):

PR #74, opened 2024-07-23
Rust/c3id: Decrease allocations & formatting

jdraaijer-microsoft:
Reducing the amount of required allocations improves performance by about 27% for few-block computations (0 to 88 characters), and by about 7% for many block (1024 characters) due to reduced (re-)allocation overhead.

[Benchmark](https://github.com/microsoft/security-utilities/blob/users/jdraaijer/bench/src/security_utilities_rust/benches/cross_company_correlating_id.rs)

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
@@ -27,8 +27,9 @@ pub fn generate_cross_company_correlating_id(text: &str) -> String {
 
 fn generate_sha256_hash<'a>(text: &str, buffer: &'a mut [u8; 64]) -> &'a [u8; 64] {
     let result = THREAD_LOCAL_SHA256.with(|sha| {
-        sha.borrow_mut().update(text.as_bytes());
-        sha.borrow_mut().finalize_reset()
+        let mut sha = sha.borrow_mut();
+        sha.update(text.as_bytes());
+        sha.finalize_reset()
     });
 
     let data: [u8; 32] = result.into();
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
@@ -4,14 +4,14 @@
 use base64::{engine::general_purpose::STANDARD, Engine as _};
 use sha2::{Digest, Sha256};
 use std::cell::RefCell;
-use std::fmt::Write;
 
 thread_local! {
     static THREAD_LOCAL_SHA256: RefCell<Sha256> = RefCell::new(Sha256::new());
 }
 
 pub fn generate_cross_company_correlating_id(text: &str) -> String {
-    let hash = generate_sha256_hash(text);
+    let mut buffer = [0u8; 64];
+    let hash = generate_sha256_hash(text, &mut buffer);
 
     let checksum = THREAD_LOCAL_SHA256.with(|sha| {
         let mut sha = sha.borrow_mut();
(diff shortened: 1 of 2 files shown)


src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
```diff
@@ -13,23 +13,25 @@ thread_local! {
...
+fn generate_sha256_hash(text: &str) -> String {
     let result = THREAD_LOCAL_SHA256.with(|sha| {
         sha.borrow_mut().update(text.as_bytes());
         sha.borrow_mut().finalize_reset()
     });
 
-    fmt::format(format_args!("{:X}", result))
-}
\ No newline at end of file
+    let mut output = String::with_capacity(2 * 32);
```
suvamM: @jdraaijer-microsoft  What's the rationale behind allocating 64 bytes? A rough upper bound? Can we store in a variable rather than hard code the value?

suvamM (dismissed):

src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
jdraaijer-microsoft: We know this value exactly because hex encoding `n` bytes will always produce `n * 2` ascii bytes. 

Now that you point it out, I actually realised we don't need this allocation either: 
1. We know all the size up front.
2. The encoding is super simple and easy to verify.

Will add a commit that removes this one too, since we're hunting for % anyways.

jdraaijer-microsoft: Done, this should now be ready to merge :)

src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
```diff
@@ -2,34 +2,55 @@
...
 
     let to_encode = &checksum[0..15];
     STANDARD.encode(to_encode)
 }
 
-pub fn generate_sha256_hash(text: &str) -> String {
-
+fn generate_sha256_hash<'a>(text: &str, buffer: &'a mut [u8; 64]) -> &'a [u8; 64] {
     let result = THREAD_LOCAL_SHA256.with(|sha| {
         sha.borrow_mut().update(text.as_bytes());
```
suvamM: @jdraaijer-microsoft This can optimized as well right?

src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
```diff
@@ -2,34 +2,55 @@
...
         sha.borrow_mut().finalize_reset()
     });
 
-    fmt::format(format_args!("{:X}", result))
-}
\ No newline at end of file
+    let data: [u8; 32] = result.into();
+
+    for (idx, byte) in data.into_iter().enumerate() {
```
suvamM, line 41: @jdraaijer-microsoft I don't entirely follow the algorithm here. Since the tests pass, I have faith in the logic. But a question: generating cross company ID should be relatively infrequent. Are the gains with this optimization high enough to trade off with respect to the readability?

src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
jdraaijer-microsoft: If it's the base64 encoding you are referring to: yes, in theory. I'm less hopeful of doing that correctly though, so I think leaving it like this is OK. Re-implementing base64 encoding was not on my bucket list, unless we think it is worthwhile.

The main reason I don't think it is useful is because we need to return a heap-allocated string anyways (and the single allocation is the only thing we could improve). `STANDARD.encode` already optimizes the allocation for the string, so we cannot really improve there.

suvamM: @jdraaijer-microsoft Referring to the double borrow happening in lines 30 and 31.

src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
jdraaijer-microsoft: It saves one allocation (~7% perf for few-block hashes). This is just a manual implementation of UpperHex formatting for the array.

I can remove it or add a clarifying comment

jdraaijer-microsoft: I've added a comment explaining what this does. IMO it's worth it since the encoding is so simple, but I can agree that it is a minor nuisance to understand

src/security_utilities_rust/src/microsoft_security_utilities_core/cross_company_correlating_id.rs
jdraaijer-microsoft: Fixed

suvamM (approved):

PR #75, opened 2024-07-23
Adding release notes for Rust crates

suvamM:
This PR creates a new release notes document for the Rust crates. This cleanly partitions updates to C# and Rust packages. The PR also moves over some entries from the existing Release Notes.

### DIFF
--- a/docs/ReleaseHistory_Rust.md
+++ b/docs/ReleaseHistory_Rust.md
@@ -1,5 +1,5 @@
 # Release Notes
-This document contains release notes pertaining to the Rust cargo packages.
+This document contains release notes pertaining to the Rust crates.
 
 ## Definitions
 
--- a/docs/ReleaseHistory_Rust.md
+++ b/docs/ReleaseHistory_Rust.md
@@ -1,0 +1,27 @@
+# Release Notes
+This document contains release notes pertaining to the Rust cargo packages.
+
+## Definitions
+
+- RUL => New detection.
+- DEP => Update dependency.
+- BRK => General breaking change.
+- BUG => General bug fix.
+- NEW => New API or feature.
+- PRF => Performance work.
+- FPS => False positive reduction in static analysis.
+- FNS => False negative reduction in static analysis.
+
+# UNRELEASED
+- DEP: `System.Text.Json` updated to `v8.0.4` to resolve Depandabot alert.
+- BRK: `Scan` struct now comprises a `ScanEngine` and a `ScanState` instance. Scan information, such as `checks`, must be accessed via the `ScanState` field of the `Scan` struct. 
+
+# 1.5.2 - 07/05/2024
+- NEW: Added an initial secret redaction capability to the Rust package.
+
+# 1.5.1 - 06/27/2024
+- DEP: Rust packages now depend on `msvc_spectre_libs` to link Spectre-mitigated libraries for `msvc` targets.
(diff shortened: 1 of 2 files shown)


jdraaijer-microsoft:
As a note, I'm not sure that the `BRK` listed here is actually breaking. See [my comment](https://github.com/microsoft/security-utilities/pull/72#discussion_r1688118473)

beaubelgrave:
> As a note, I'm not sure that the `BRK` listed here is actually breaking. See [my comment](https://github.com/microsoft/security-utilities/pull/72#discussion_r1688118473)

Agree, useful to have without the BRK though.

suvamM:
> As a note, I'm not sure that the `BRK` listed here is actually breaking. See [my comment](https://github.com/microsoft/security-utilities/pull/72#discussion_r1688118473)

@jdraaijer-microsoft @beaubelgrave Can you take a look at the updated entry?

@michaelcfanning You're right---`NEW` is more appropriate. Updated.

docs/ReleaseHistory_Rust.md
```diff
@@ -0,0 +1,27 @@
...
+- BRK => General breaking change.
+- BUG => General bug fix.
+- NEW => New API or feature.
+- PRF => Performance work.
+- FPS => False positive reduction in static analysis.
+- FNS => False negative reduction in static analysis.
+
+# UNRELEASED
+- DEP: `System.Text.Json` updated to `v8.0.4` to resolve Depandabot alert.
+- NEW: Introduces the `ScanEngine` struct, which now encapsulates the internal state tracking previously managed by the `Scan` struct. This refactor simplifies the usage of the `Scan` struct in scoped thread contexts, removing the requirement for a `Mutex` or `Arc`.
```
beaubelgrave: You have this kind of backwards :)

The ScanEngine is the base, that is entirely immutable and takes a single mutable ScanState. This makes it possible for concurrent usage.

The Scan struct is unchanged from the customer perspective, however, it now leverages an internal ScanState and ScanEngine to ensure no-breaking changes.

beaubelgrave: Customers wishing to use concurrency without any locking should use a single ScanEngine instance with ScanState's per-invocation/thread.

suvamM: @beaubelgrave I think I understand your point. Took another stab at authoring the note.

beaubelgrave: > This refactor simplifies the usage of the `Scan` struct in scoped thread contexts, removing the requirement for a `Mutex` or `Arc`.

This should change to:
This refactor allows the usage of a single `ScanEngine` struct in scoped thread contexts, removing the requirement for a `Mutex` or `Arc`.

suvamM: @beaubelgrave I think you replied on an outdated version of the note. I updated it to: 

Introduces the `ScanEngine` struct, which allows simplified usage in concurrent scenarios---a single `ScanEngine` instance, along with per-thread `ScanState` instances, suffice without the need for additional synchronization. The existing `Scan` struct is operationally unchanged for users.

beaubelgrave: Perfect! Let's wait for a release until we get the Rc's removed, I think that's the last step to full concurrency.

suvamM: Sounds good! Thanks @beaubelgrave .

PR #77, opened 2024-07-24
Rust/His: Make `ScanEngine` `Send` and `Sync`

### LINKED ISSUE #76: Rust: `ScanEngine` is `!Sync` and `!Send`
Because `ScanEngine` includes an `Rc` and a non-send-and-sync `dyn Fn`, it is neither `Send` nor `Sync`. This means that re-using it in multi-threaded applications is not (trivially) possible. 

This makes it so neither of the following compile: 

```rust
    const fn assert_send<T: Send>() {}
    const fn assert_sync<T: Sync>() {}

    // These do not compile :(
    const _: () = assert_sync::<ScanEngine>();
    const _: () = assert_send::<ScanEngine>();

    let engine = ScanEngine::new(Default::default());
    std::thread::scope(|s| {
        s.spawn(|| {
            let mut state = ScanState::default();
            engine.byte_scan(&mut state, &[]);
        });
    });

    let engine = ScanEngine::new(Default::default());
    std::thread::spawn(move || {
        let mut state = ScanState::default();
        engine.byte_scan(&mut state, &[]);
    });
```

To fix it, we can: 
1. Tighten the bound on all `validator`s to in include `+ Send + Sync` and replace the `Rc` with an `Arc`.
2. Replace the `Rc` with a function pointer outright (i.e. `fn(&[u8]) -> usize`).

The effect of 1 is that the validator is a slightly less flexible, but for most practical applications this should be fine. All of the definitions in the project currently are effectively `fn(&[u8]) -> usize`, so they strictly adhere to it. Most other validators should also own their data and therefore adhere to the new bound, so impact should be minimal.

For 2, it means that the validators become, effectively, compile-time variables that can not change at runtime. That sounds undesirable, but I wanted to highlight it as a potential solution.

[0]: https://doc.rust-lang.org/std/sync/struct.Arc.html

jdraaijer-microsoft:
Closes #76 

See issue for more detailed description.

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-use std::rc::Rc;
+use std::sync::Arc;
 
 /* Indicates the char is part of a small mask */
 const MASK_SMALL: u8 = 1 << 0;
@@ -109,7 +109,7 @@ pub struct PossibleScanMatch {
     start: u64,
     len: usize,
     utf8: bool,
-    validator: Rc<dyn Fn(&[u8]) -> usize>,
+    validator: Arc<dyn Fn(&[u8]) -> usize + Send + Sync>,
 }
 
 impl PossibleScanMatch {
@@ -119,7 +119,7 @@ impl PossibleScanMatch {
         start: u64,
         len: usize,
         utf8: bool,
-        validator: Rc<dyn Fn(&[u8]) -> usize>) -> Self {
+        validator: Arc<dyn Fn(&[u8]) -> usize + Send + Sync>) -> Self {
         Self {
             name,
             def_index,
@@ -257,7 +257,7 @@ pub struct ScanDefinition {
     mask_size: u8,
     packed_utf8: u64,
     packed_utf16: u64,
-    validator: Rc<dyn Fn(&[u8]) -> usize>,
+    validator: Arc<dyn Fn(&[u8]) -> usize + Send + Sync>,
 }
 
 impl ScanDefinition {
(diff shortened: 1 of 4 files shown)


src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-use std::rc::Rc;
+use std::sync::Arc;
```
beaubelgrave, line 4: Can we just remove these fully? I believe since the validator is just fn you can do it, right?

jdraaijer-microsoft: We can, but then it also prevents anyone else from implementing validators that are not compile-time constant (although I suppose just using a static Mutex would actually solve this problem, so perhaps not as bad of a problem as I imagine it being). I guess we could always loosen the constraint later in the case that it turns out that we wish to support that.

beaubelgrave: Interesting, good point. @michaelcfanning likely owns that decision.

jdraaijer-microsoft: Will ~~do~~ wait ;)!

jdraaijer-microsoft: Have left it as-is for now, with a changelog entry. Let's see what Michael says. 

I'm in EU timezones, so will not be available for some time, now.

michaelcfanning: I think in general it's fine for this library to be in a bit of flux. A tighter constraint that we loosen later is easy to handle from a compat standpoint than trying to move the other direction.

If Beau doesn't have a strong feeling, let's get this in and we can revisit as Johannes points out later if it comes up.

michaelcfanning (approved):
:shipit:

PR #79, opened 2024-07-31
Rust: add benchmarks

jdraaijer-microsoft:
Add some benchmarks, using the [`Criterion`][0] framework.

The benchmarks added in this PR are basic, and the ones for more advanced usage probably need better real-world data that would give us somewhat more accurate information about real-life performance.

[0]: https://docs.rs/criterion/latest/criterion/


src/security_utilities_rust/benches/identifiable_scans.rs
```diff
@@ -0,0 +1,76 @@
...
+criterion_group! {
+    name = id_scans;
+    config = Criterion::default().warm_up_time(Duration::from_millis(500));
+    targets = id_scans_benchmark,
+}
+
+criterion_main!(id_scans);
+
+/*
+pub fn write_data() {
```
suvamM, line 30: Can this commented code be removed?

suvamM (approved):

PR #82, opened 2024-08-05
Rust/His: Add missing scan_defs() method on Scan

### LINKED ISSUE #81: Rust ffi crate breaks due to recent changes to core
Recent changes to `security_utilities_rust` causes build breaks in `security_utilities_rust_ffi`. This causes the downstream C# build to break as well.

To do: update `security_utility_rust_ffi` to support the new APIs in `security_utility_rust`.

beaubelgrave:
When the Scan struct was updated to use a ScanEngine, the scan_defs() method was missed being added. This caused the FFI crate to break unknowingly.

Add scan_defs() to the Scan struct, so it behaviors exactly as before.

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
@@ -1128,6 +1128,8 @@ impl Scan {
 
     pub fn possible_matches(&self) -> &Vec<PossibleScanMatch> { self.state.possible_matches() }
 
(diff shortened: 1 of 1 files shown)


michaelcfanning (approved):
What change can we make to have this kind of problem break our check-in process? It looks like BuildAndTest.cmd broke here, are we running this script and checking error level on return in our CI/CD infra?

@suvamM may have some insight. I'm glad to merge this immediate change as it will unblock separate stuff going on.

michaelcfanning:
And just FYI, I have now verified that BuildAndTest.cmd now succeeds for me against a clean main...

PR #83, opened 2024-08-06
New API to make it more convenient to use byte[] args.

michaelcfanning:
* NEW: Provide `StandardCommonAnnotatedKeySizeInBytes` and `LongFormCommonAnnotatedKeySizeInBytes` constants (63 and 64, respectively).
* NEW: `TryValidateCommonAnnotatedKey(byte[], string)` to facilitate working with keys as byte arrays.
* NEW: `ComputeDerivedCommonAnnotatedKey(string, byte[])` to facilitate working with keys as byte arrays.
* NEW: `GenerateCommonAnnotatedKeyBytes(bool, byte[], byte[], bool, char?)` to facilitate working with keys as byte arrays.
* NEW: Change `ComputeDerivedCommonAnnotatedKey(string, string)` `textToHash` parameter name to `derivationInput` to better reflect its purpose.

Added the release notes legend to the document. These API were requested by a project team interested in using keys strictly as byte[]. The API are not yet tuned for performance, i.e., they most simply offload work to the text-based API. This is required due to textual matching that's in place for verifying the validity of generated key character set, etc. A more tuned byte[]-based API would have to examine the data differently to perform validity checks such as these.


src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -59,6 +64,40 @@ public static bool IsBase64UrlEncodingChar(this char ch)
...
+
+        bool longForm = key.Length == LongFormCommonAnnotatedKeySizeInBytes;
+
+        ulong checksumSeed = VersionTwoChecksumSeed;
+
+        byte[] bytesToChecksum = new byte[60];
+        Array.Copy(key, bytesToChecksum, bytesToChecksum.Length);
+
+        
+        int checksum = Marvin.ComputeHash32(bytesToChecksum, checksumSeed, 0, bytesToChecksum.Length);
```
suvamM, line 83: Consider using the LINQ `key.Take(60)`? Then the array copying in lines 80-81 can go away. Also makes the function call "Rust-like".

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -59,6 +64,40 @@ public static bool IsBase64UrlEncodingChar(this char ch)
...
+
+        byte[] checksumBytes = BitConverter.GetBytes(checksum);
+
+        int firstChecksumByteIndex = 60;
```
suvamM: Consider defining this variable earlier. Then we don't need the constant in line 80.

suvamM (dismissed):

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
michaelcfanning: this apparently simply change is actually disruptive because other API work directly against byte[] rather than IEnumerable<byte>. that probably should be changed but it's API breaking so not doing it now. :)

The code you proposed to change is just a one-liner, so I think it's ok to retain, though arguably not as readable/elegant as your suggestion.

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
michaelcfanning: Good one, thanks.

ReleaseHistory.md
```diff
@@ -1,4 +1,26 @@
...
+- FPS => FP reduction in static analysis
+- FNS => false negative reduction in static analysis
+- FPD => FP reduction in dynamic phase
+- FND => False negative reduction in dynamic phase
+- UER => eliminate unhandled exceptions in rules
+- UEE => eliminate unhandled exceptions in engine
+- DEP => upgrade dependency versions
+- NEW => new feature 
+
+UNRELEASED
```
suvamM: @michaelcfanning We have `docs/ReleaseNotes.md`. Is the plan to abandon that and maintain updates in `ReleaseHistory`?

PR #88, opened 2024-08-12
Adding portable pdb generation support for Utilities.core

suvamM:


src/Microsoft.Security.Utilities.Core/Microsoft.Security.Utilities.Core.csproj
```diff
@@ -18,6 +18,10 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <PropertyGroup>
```
michaelcfanning: >PropertyGroup [](http://example.com/codeflow?start=3&length=13)

shouldn't we put this in build.props? it seems more likely to be picked up by upstream builds...

suvamM: @michaelcfanning That makes sense. Good point.

shaopeng-gh: for my understanding this is to change format form `default` to `portable`,
not `Adding portable pdb generation support`, 
the `default` is different for the .net version used.
Do we think the default was `Full` ?

suvamM: @shaopeng-gh Per [this](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation), the default format does seem to be `full`.

shaopeng-gh: for vscode wiki it says: `With .NET Core "SDK"-style .csproj's, Portable PDBs are already enabled by default` 
[https://github.com/dotnet/vscode-csharp/wiki/Portable-PDBs](https://github.com/dotnet/vscode-csharp/wiki/Portable-PDBs)

// First four bytes should be 'BSJB'
[https://github.com/microsoft/vstest/blob/80af388e851d0d8b88f93d07d46d47054acb9951/src/Microsoft.TestPlatform.ObjectModel/Navigation/PortablePdbReader.cs#L106](https://github.com/microsoft/vstest/blob/80af388e851d0d8b88f93d07d46d47054acb9951/src/Microsoft.TestPlatform.ObjectModel/Navigation/PortablePdbReader.cs#L106)

it looks like already is in my local build even in net451 folder:
`\security-utilities\bld\bin\AnyCPU_Release\Microsoft.Security.Utilities.Core\net451\Microsoft.Security.Utilities.Core.pdb`

image

suvamM: @shaopeng-gh This is in response to the following error message I received while publishing the package to Nuget:

image

This is why I am trying out setting the `portable` DebugType explicitly. You have reasons to believe this won't help?

shaopeng-gh: The one that is not with `BSJB` is the  lower case one. 
Is that a ref, ref will not get changed by the settings in current repo.

image

PR #92, opened 2024-09-09
Api updates to unify all key creation through a common helper.

michaelcfanning:
- BRK: Rename `StandardCommonAnnotatedKeySize` to `StandardEncodedCommonAnnotatedKeySize` and `LongFormCommonAnnotatedKeySize` to `LongFormEncodedCommonAnnotatedKeySize` to distinguish these from const values for key lengths in bytes.
- BUG: Correct `CommonAnnotatedKeyRegexPattern` to detect keys (as denoted by `H` in the platform signature) derived from hashing data with CASK keys or arbitrary secrets.
- BUG: Fix issue in low-level `GenerateCommonAnnotatedTestKey` helper in which key kind signature was hard-coded for `D` (derived) for both derived and hashed keys (which should be denoted by `H`).
- NEW: Add `ComputeCommonAnnotatedHash` to generate annotated fingerprints from arbitrary strings.
- NEW: Add `CommonAnnotatedDerivedKeySignature` and `CommonAnnotatedHashedDataSignature` to denote these generated key variations.
- NEW: Update key generation to use Base62 for all encoded checksums (including primary keys). As a result, all test keys (in which the randomized component is a common character) will be valid (because we no longer will generate special characters in the computed checksum).
- NEW: Add `longForm` argument to `ComputeDerivedCommonAnnotatedKey` and `ComputeCommonAnnotatedHash` to support backwards-compatible, full 64-byte encoded forms of these keys.
- NEW: Provide `ComputeDerivedCommonAnnotatedKey` and `ComputeCommonAnnotatedHash` overloads that accept an arbitrary secret (and which allow platform and provider data to be explicitly specified).

Previously, we appear to have shipped `1.6.0` without including the `ComputeCommonAnnotatedHash` API. This API as authored required a CASK secret to operate but this is clearly not sufficient: providers that update their logic for producing derived keys and hashes may need to operate against a legacy secret that drives the signing process. To add this capability, however, we need to update the API to accept all the data (i.e., the CASK reserved platform + provider data) that is possible to encoded in a key. This was not previously required because we flowed that data from the CASK signing secret. This has the added benefit of allowing users to decide whether to flow data from a primary CASK key or to override it.

While making this change, I decided to take the plunge and simply make our key generation logic use a Base62 checksum. One positive result is that every test key (by CASK specification) will generate a legal key (because these specified patterns will *never* produce a special character that's illegal in CASK). We already had to use Base62 for the derived and hashed key cases; it is not a stretch to do so for primary keys. I've left back-up logic in the API for handling the case where the checksum is expressed as decoded base64 bytes. That means this library will continue to handle any historical key minters that are producing the older checksums.

This work also revealed a couple of small bugs related to the new hashing API: we needed to update our common regex to understand the `H` reserved key kind denoting this type and I fixed an issue in the literal generation logic that was not properly encoding `D` for derived keys and `H` for hashed.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
@@ -18,13 +18,13 @@ namespace Microsoft.Security.Utilities
             ulong checksumSeed = IdentifiableSecrets.VersionTwoChecksumSeed;
             string base64EncodedSignature = key.Substring(76, 4);
 
-            if (key.Length != IdentifiableSecrets.StandardCommonAnnotatedKeySize &&
-                key.Length != IdentifiableSecrets.LongFormCommonAnnotatedKeySize)
+            if (key.Length != IdentifiableSecrets.StandardEncodedCommonAnnotatedKeySize &&
+                key.Length != IdentifiableSecrets.LongFormEncodedCommonAnnotatedKeySize)
             {
                 return false;
             }
 
-            bool longForm = key.Length == IdentifiableSecrets.LongFormCommonAnnotatedKeySize;
+            bool longForm = key.Length == IdentifiableSecrets.LongFormEncodedCommonAnnotatedKeySize;
 
             // This code path is intended to ensure that common annotated security keys are
             // highly backwards compatible with the older identifiable keys format. This
--- a/src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
@@ -17,7 +17,6 @@ namespace Microsoft.Security.Utilities
             secret = null;
             ulong checksumSeed = IdentifiableSecrets.VersionTwoChecksumSeed;
             string base64EncodedSignature = key.Substring(76, 4);
-            bool isDerived = key[DerivedKeyCharacterOffset] == 'D';
 
             if (key.Length != IdentifiableSecrets.StandardCommonAnnotatedKeySize &&
                 key.Length != IdentifiableSecrets.LongFormCommonAnnotatedKeySize)
@@ -42,11 +41,15 @@ namespace Microsoft.Security.Utilities
                 int checksum = Marvin.ComputeHash32(keyBytes, checksumSeed, 0, keyBytes.Length - 4);
 
                 byte[] checksumBytes = BitConverter.GetBytes(checksum);
-                string encodedChecksum = isDerived ? checksum.ToBase62() : Convert.ToBase64String(checksumBytes);
+                string encodedChecksum = checksumBytes.ToBase62();
 
                 if (!encodedChecksum.StartsWith(partialEncodedChecksum))
                 {
-                    return false;
+                    encodedChecksum = Convert.ToBase64String(checksumBytes);
+                    if (!encodedChecksum.StartsWith(partialEncodedChecksum))
+                    {
+                        return false;
+                    }
                 }
 
                 keyBytes[keyBytes.Length - 4] = checksumBytes[0];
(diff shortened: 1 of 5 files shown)


src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
```diff
@@ -42,11 +41,15 @@ public static bool TryCreate(string key, out CommonAnnotatedKey secret)
                 int checksum = Marvin.ComputeHash32(keyBytes, checksumSeed, 0, keyBytes.Length - 4);
 
                 byte[] checksumBytes = BitConverter.GetBytes(checksum);
-                string encodedChecksum = isDerived ? checksum.ToBase62() : Convert.ToBase64String(checksumBytes);
+                string encodedChecksum = checksumBytes.ToBase62();
```
rwoll, line 44: Does this mean old validation code will no longer work?

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -81,18 +85,26 @@ public static bool TryValidateCommonAnnotatedKey(byte[] key,
         Array.Copy(key, bytesToChecksum, bytesToChecksum.Length);
         
         int checksum = Marvin.ComputeHash32(bytesToChecksum, checksumSeed, 0, bytesToChecksum.Length);
```
rwoll, line 87: Looking at the changes in these methods, does this imply `TryValidateCommonAnnotatedKey` will no longer validate keys minted prior to this change?

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -258,26 +258,39 @@ public void IdentifiableSecrets_GeneratedDerivedAndHashedAnnotatedKeys()
...
+                            string computedKey = derived
+                                ? IdentifiableSecrets.ComputeDerivedCommonAnnotatedKey(textToHash, key, longForm)
+                                : IdentifiableSecrets.ComputeCommonAnnotatedHash(textToHash, key, longForm);
+
+                            string label = derived ? "derived" : "hashed";
+
+                            bool result = CommonAnnotatedKey.TryCreate(computedKey, out CommonAnnotatedKey caKey);
+                            result.Should().BeTrue(because: $"the derived key '{computedKey}' should be a valid common annotated security key");
+
+                            bool keyKindCorrect = derived ? caKey.IsDerivedKey : caKey.IsHashedDataKey;
```
rwoll, line 272: This test logic has high cognitive load.

rwoll: I understand the goal of trying to mash a bunch of test cases in a `for loop`, but breaking them out and having explicit test names and expectations will make them easier to follow for reviewers.

Consider unfolding this test (and others) into explicit cases and expectations without computing the latter on the fly.

rwoll (commented):
Reviewed but would like to go over this in more detail. Let's also get @suvamM on the call to ensure multiple eyes.

A meta comment: the test cases with for loops and computed expectations are "too smart". Please hard code the expecations and don't derive it from nested loop variables.

src/Microsoft.Security.Utilities.Core/CommonAnnotatedKey.cs
michaelcfanning: As discussed, old checksums *will* work but we require a directed test which I will add in a follow on change.

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
michaelcfanning: Nope, works, but I should add a specific test to prove this. :)

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
michaelcfanning: I completely understand. One thing about a test approach like this, the debuggability experience is non-obvious. i.e., though it's clearly not easy code to read, it's designed to be extremely easy to *support* when it breaks. You'd have to provoke and observe the assertions output to form your own opinion.

I like these super-powered tests, that comprehensively exercise something in a very compact form. However, there are downsides as you've observed and, in my opinion, these tests s/be supplemented with more focused, directed tests.

Another advantage of my approach is that it avoids lots of duplicated code that ends up being fragile/requiring lots of updating when things change. By building this kind of test and then authoring very granular directed tests, we may be able to have the best of both worlds. Ideally, any time this test breaks: 1) its output makes it extremely easy to debug (maybe I'll take a moment to demo the xp), 2) the break of the core test should always occur in sync (ideally) with a more atomic unit test that gets to the underlying variant/whatever that's been broken.

Good conversation to be continued.

rwoll (approved):
Stamped with offline caveats.

PR #93, opened 2024-09-12
Provide explicit confidence level for missing rules.

michaelcfanning:
- BUG: Mark `SEC000/000.Unclassified32ByteBase64String`, `SEC000/001.Unclassified64ByteBase64String`, `SEC101/101.AadClientAppLegacyCredentials`, `SEC000/001.Unclassified64ByteBase64String` as `DetectionMetadata.LowConfidence`.
- BUG: Mark `SEC101/109.AzureContainerRegistryLegacyKey` as `DetectionMetadata.MediumConfidence`.
- BUG: Mark `SEC101/030.NuGetApiKey`, `SEC101/105.AzureMessageLegacyCredentials`, `SEC101/110.AzureDatabricksPat`,`SEC101/050.NpmAuthorKey`,`SEC101/565.SecretScanningSampleToken` as `DetectionMetadata.HighConfidence`.
- PRF: Enable scan pre-filtering by declaring `.servicebus` as `SEC101/105.AzureMessageLegacyCredentials` signature.

I've introduced a new unit-test that forces a development invariant, that every pattern include an explicit confidence (i.e., accuracy or precision) designation, between `High`, `Medium` and `Low`. I've also updated doc comments to try to describe these categories. Maybe I should include this as well but typically:

- Low == implemented only on an ad hoc basis for deep research or review purposes.
- Medium == suitable to surface as warnings and/or bypassable findings in blocking scenarios.
- High == very low noise rates, could be candidate for hard-blocking scenarios.
- Identifiable == mathematically nearly certain to be a true positive. Suitable for hard-blocking/other prescriptive controls.

While doing this work I noticed a medium confidence pattern was missing a signature used for pre-filtering (in .NET, these values are confirmed to present in a scan target using `string.IndexOf`, a check which is empirically observed to be ~10x - 20x faster than non-back-tracking regex engines such as `RE2`. I've added a second invariant test that insists all patterns of medium confidence or higher include one or more signatures to enable the prec-check.

@nguerrera @suvamM @rwoll @shaopeng-gh @yongyan-gh @LingZhou-gh @evelyn-ys

### DIFF
--- a/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
@@ -318,6 +318,24 @@
     "DetectionMetadata": "Identifiable"
   },
   {
+    "Pattern": "(^|[^0-9a-z])(?<refine>oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4])([^aeimquy4]|$)",
+    "Id": "SEC101/030",
+    "Name": "NuGetApiKey",
+    "Signatures": [
+      "oy2"
+    ],
+    "DetectionMetadata": "HighEntropy, FixedSignature, HighConfidence"
+  },
+  {
+    "Pattern": "(?:^|[^0-9a-f\\-])(?P<refine>dapi[0-9a-f\\-]{32,34})(?:[^0-9a-f\\-]|$)",
+    "Id": "SEC101/110",
+    "Name": "AzureDatabricksPat",
+    "Signatures": [
+      "dapi"
+    ],
+    "DetectionMetadata": "HighEntropy, HighConfidence"
+  },
+  {
     "Signatures": [
       "AZEG"
     ],
@@ -331,5 +349,23 @@
     "Id": "SEC101/190",
     "Name": "AzureEventGridIdentifiableKey",
     "DetectionMetadata": "Identifiable"
+  },
+  {
+    "Pattern": "(^|[^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890])(?<refine>npm_[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{36})([^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]|$)",
+    "Id": "SEC101/050",
+    "Name": "NpmAuthorKey",
+    "Signatures": [
+      "npm_"
+    ],
+    "DetectionMetadata": "HighEntropy, FixedSignature, HighConfidence"
+  },
+  {
+    "Pattern": "(^|[^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890])(?P<secret>secret_scanning_ab85fc6f8d7638cf1c11da812da308d43_[0-9A-Za-z]{5})([^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]|$)",
+    "Id": "SEC101/565",
+    "Name": "SecretScanningSampleToken",
+    "Signatures": [
+      "ab85"
+    ],
+    "DetectionMetadata": "HighEntropy, FixedSignature, HighConfidence"
   }
 ]
--- a/GeneratedRegexPatterns/LowConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/LowConfidenceSecurityModels.json
@@ -8,5 +8,40 @@
       "ret="
     ],
     "DetectionMetadata": "LowConfidence"
+  },
+  {
+    "Pattern": "^(?i)[a-z0-9.=\\-:[_@\\/*\\]+?]{32}$",
+    "Id": "SEC000/003",
+    "Name": "Unclassified32CharacterString",
+    "Signatures": null,
+    "DetectionMetadata": "HighEntropy, Unclassified, LowConfidence"
+  },
+  {
+    "Pattern": "^[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890+/]{43}=$",
+    "Id": "SEC000/000",
+    "Name": "Unclassified32ByteBase64String",
+    "Signatures": null,
+    "DetectionMetadata": "HighEntropy, Unclassified, LowConfidence"
+  },
+  {
+    "Pattern": "^[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890+/]{86}==$",
+    "Id": "SEC000/001",
+    "Name": "Unclassified64ByteBase64String",
+    "Signatures": null,
+    "DetectionMetadata": "HighEntropy, Unclassified, LowConfidence"
+  },
+  {
+    "Pattern": "^[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890\\-_~.]{34}$",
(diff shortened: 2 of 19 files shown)


GeneratedRegexPatterns/HighConfidenceSecurityModels.json
```diff
@@ -317,6 +317,24 @@
     "Name": "AzureContainerRegistryIdentifiableKey",
     "DetectionMetadata": "Identifiable"
   },
+  {
+    "Pattern": "(^|[^0-9a-z])(?<refine>oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4])([^aeimquy4]|$)",
+    "Id": "SEC101/030",
+    "Name": "NuGetApiKey",
```
michaelcfanning, line 323: >NuGetApiKey [](http://example.com/codeflow?start=13&length=11)

It appears these patterns were added without regen'ing the rules. We should make the build script break if there are open edits after it executes (which could indicate this auto-generated file was updated and needs to be checked in).

GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json
```diff
@@ -45,34 +45,34 @@
     "Id": "SEC000/003",
     "Name": "Unclassified32CharacterString",
     "Signatures": null,
-    "DetectionMetadata": "ObsoleteFormat, HighEntropy, Unclassified"
```
michaelcfanning, line 48: >ObsoleteFormat [](http://example.com/codeflow?start=26&length=14)

This pattern *used* to cover AAD 32 character passwords, which are obsolete in format. We no longer actually cover this legacy pattern, which is five generations out as a format, extremely noisy, etc.

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
```diff
@@ -10,6 +17,16 @@
...
   },
+  {
+    "Pattern": "https?:\\/\\/(?:[^:@]+):(?<refine>[^:@?]+)@",
+    "Id": "SEC101/127",
+    "Name": "UrlCredentials",
+    "Signatures": [
+      "http",
+      "https"
+    ],
+    "DetectionMetadata": "MediumConfidence"
```
michaelcfanning, line 28: >MediumConfidence [](http://example.com/codeflow?start=26&length=16)

I am not strictly sure this one is Medium.

src/Microsoft.Security.Utilities.Core/DetectionMetadata.cs
```diff
@@ -18,15 +18,47 @@ public enum DetectionMetadata
...
+    /// appropriate in scenarios where the productivity costs of reviewing and ignoreing
+    /// false positives is offset by the security value achieved. A scan system that
+    /// identifies medium confidence findings in incremental code changes, for example, can
+    /// typically amortize the productivity costs of medium confidence findings. A system
+    /// that attempts to hard-block processes based on detecting an apparent credential
+    /// would not enable medium confidence patterns.
+    /// </summary>
     MediumConfidence = 1 << 7,
 
+    // The confidence that this pattern comprises sensitive data is very high. i.e., the
```
suvamM, line 59: The distinction between `HighConfidence` and `Identifiable` is rather ambiguous to me. `HighConfidence` has a "very high confidence", while "Identifiable" confidence is "effectively 100%"---very subjective. We should either 
1. define mathematical limits on detection accuracy that separates the two groups, or
2. define syntactic features which separates the two (for example: "for a find to qualify as identifiable, it must have a fixed signature, embedded checksum, etc").

The same logic extends to the other classifications. When does a category stop being medium confidence, and transition to high confidence? We need some objective metrics to define these classifications. Not a blocker for this PR, but should be something we think about.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -25,6 +26,84 @@ public class WellKnownRegexPatternsTests
             "SEC101/109.AzureContainerRegistryLegacyKey"
         };
 
+        [TestMethod]
+        public void WellKnownRegexPatterns_EnsureAllPatternsExpressConfidence()
```
suvamM, line 30: What a fantastic test!

suvamM (approved):

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
suvamM: Ties into my other comment: we don't have an objective metric to define the confidence levels.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: :) This is how I find the things I needed to fix. I love this approach because:

1) you codify a development invariant that guarantees future contributors do not regress in an important concern
2) by authoring the test to drive resolving the gap, you guarantee that the test output itself is very actionable (to help those future devs who aren't aware of the standard).

rinse and repeat for all the quality bars you want to enforce in your plug-ins.

rwoll: Screenshot 2024-09-12 at 9 06 28 AM

^ It looks like the C# bits of this project do not run with each PR like we've setup Rust and Node. This test will still eventually run but consider adding CI steps similar to Rust + Node that we have in this repo to catch the issues earlier (this will also help catch the lack of checking in the latest generated code).

rwoll (approved):
Overall, LGTM. @suvamM's comments on distinction b/w the confidence levels resonate with me, but reading your descriptions, it sounds like the decision tree is:

If, formally `Identifiable` cred, mark it as that.
If very noisy mark it as `Low`,
If rarely noisy but not `Identifiable`, mark it as `High`,
If none of the above, mark as `Medium`?

Does that sound correct?

src/Microsoft.Security.Utilities.Core/DetectionMetadata.cs
michaelcfanning: I agree with all you , but there is an intractable issue here, which is that we can't quantify noise levels because we don't control all data generation in scan targets. Imagine a fixed signature of `MSFT`. injected into an apparent base64-encoded string, that signature provides  ~1 in 16M odds of not being our injected sig (i.e., it was generated randomly). The problem is that not every pattern we might be inspecting is randomly generated data that has been base64-encoded. `MSFT` is a pattern that is guaranteed to be prevalent in other non-randomized data generators, like any text that is referencing the stock ticker. In other cases a prefix (like `dapi_` for databricks PATS) happens to apparently never conflict with other patterns in engineering artifacts. But this isn't for mathematical reasons.

The identifiable techniques protect against this by adding the checksum as a protection against some bad break where a prefix matches a common natural language term or whatever.  In concert with a fixed signature our original identifiable technique drive noise rates to 1 in quadrillions of matches. The latest revision goes into the quintillions. That's enough for sure. 

I have come to think that the 'powerball' test is helpful. Once you get false positive rates into the 1 in hundreds of millions domain, the odds of a false positive are the same as winning the powerball for any particular drawing. Those odds, I think, are probably a decent threshold to keep in mind (and this analogy is very understandable to humans).

Anyway, I took note of this comment and Ross', let's keep digging in (and come back to this code with improved code comments, etc., based on that discussion).

GeneratedRegexPatterns/LowConfidenceSecurityModels.json
```diff
@@ -8,5 +8,40 @@
       "ret="
     ],
     "DetectionMetadata": "LowConfidence"
+  },
+  {
+    "Pattern": "^(?i)[a-z0-9.=\\-:[_@\\/*\\]+?]{32}$",
```
evelyn-ys, line 13: Can we change the pattern to `(?i)^[a-z0-9.=\\-:[_@\\/*\\]+?]{32}$`?

I got `re.error: global flags not at the start of the expression` while testing

PR #94, opened 2024-09-12
Ensure that CASK secrets round-trip through base64 without changing.

michaelcfanning:
- BUG: Make round-tripping of common annotated security keys through base64 encoding/decoding more robust. We previously emitted illegal ending base64 characters (when appending base62 encoded checksums).

When generating CASK keys, we previously would emit final characters in the apparent base64 encoded string that weren't actually legal as the final encoded character (which must be constrained to 2 bits, one of the `A`, `Q`, `g` or `w` characters.

.NET is very forgiving of this issue when decoding and re-encoding, it will simply stop processing at the appropriate byte boundary and halt. As a result, a round-tripped CASK secret might differ from the original CASK that was emitted.

Now, the round-tripping to ensure base64-validity is added to the key minting logic.

As @rwoll has pointed out, the base64 vs. base62 confusion has resulted in unfortunate complexity in emitting this standard. Before finalizing it as a public format, we should fix that. It's too late for existing adopters! Such is the tyranny of back-compat.

Still, as long as we maintain compatibility with legacy keys, we *can* design (yet another) revision.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -435,7 +435,7 @@ public static class IdentifiableSecrets
         byte[] checksumBytes = BitConverter.GetBytes(checksum);
         string checksumText = GetBase62EncodedChecksum(checksumBytes);
 
-        key = $"{Convert.ToBase64String(keyBytes).Substring(0, 80)}{checksumText}==";
+        key = $"{Convert.ToBase64String(keyBytes).Substring(0, 80)}{checksumText}";
 
         if (!longForm)
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -445,10 +445,12 @@ public static string GenerateCommonAnnotatedTestKey(byte[] randomBytes,
...
     }
 
-    private static string GetBase62EncodedChecksum(byte[] checksumBytes)
+    internal static string GetBase62EncodedChecksum(byte[] checksumBytes)
     {
         string checksumText = checksumBytes.ToBase62();
-        checksumText = $"{checksumText}{new string('0', 6 - checksumText.Length)}";
+        checksumText = $"{checksumText}{new string('0', 6 - checksumText.Length)}==";
+
+        checksumText = Convert.ToBase64String(Convert.FromBase64String(checksumText));
```
rwoll, line 453: Is this to ensure it roundtrips?

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -32,6 +34,43 @@ static IdentifiableSecretsTests()
...
+        {
+            int failed = 0;
+            int succeeded = 0;
+            int iterations = 1000;
+
+            for (int i = 0; i < iterations; i++)
+            {
+                foreach (bool longForm in new[] { true, false })
+                {
+                    string testKey = IdentifiableSecrets.GenerateCommonAnnotatedKey("TEST",
```
rwoll, line 48: Also try with non-test keys? i.e. use `TEST` sig, but also randomize this value to cover your bases.

rwoll (approved):
Can you comment this code liberrally? Just reading the code, it's hard to track/understand the Base62 v. Base64 games and magic lengths and padding.

Issue #95, opened 2024-09-15
Regex rules double fire on some generated patterns.


michaelcfanning:
On surface inspection, we're not requiring discrete patterns for some checks, like `SEC101/101`

PR #96, opened 2024-09-15
Byte array corrections

michaelcfanning:
- BUG: Correct `IdentifiableSecrets` `ComputeDerivedCommonAnnotatedKey` and `ComputeCommonAnnotatedHash` helpers to preserve all randomized byte input entropy by encoding and decoding this data as base64.
- NEW: Add `CommonAnnotatedKey` `ChecksumBytes` and `ChecksumBytesIndex` convenience methods for retrieving key checksum data.

This change ensures that all byte[] randomized data is processed as base64 (rather than encoding and decoding as UTF8 a process that stripped entropy). This change includes testing that achieves effectively 100% code coverage (which will assist in future refactoring/simplification of code). 

Regrettably, I also noted that the `SecretMasker` test class had been commented out. As a result, tests have not been running for some period of time. I've fixed up some regressions that have accumulated but did not fix a condition in which some lower confidence findings produce duplicated results for the same test data. #95 opened to address this.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -80,31 +80,34 @@ public static class IdentifiableSecrets
 
         ulong checksumSeed = VersionTwoChecksumSeed;
 
-        int firstChecksumByteIndex = 60;
+        int firstChecksumByteIndex = CommonAnnotatedKey.ChecksumBytesIndex;
         byte[] bytesToChecksum = new byte[firstChecksumByteIndex];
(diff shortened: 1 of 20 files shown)


src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -360,12 +363,6 @@ public static string GenerateCommonAnnotatedTestKey(byte[] randomBytes,
             }
 
             key = keyBytes.ToBase62();
-
-            if (key.Length < 86)
```
michaelcfanning, line 364: >key [](http://example.com/codeflow?start=16&length=3)

These constraints are now built-in/inherent to code logic.

cfaucon (approved):

PR #101, opened 2024-10-23
Fix NuGetApiKey detection to not require Guid bytes

joelverhagen:
This also simplifies the regex to just expect oy2 and then 43 lowercase base32 characters.

Based on internal feedback we can improve the implementation to:
- Not use Guid.NewGuid for CPRNG
- Allow zero bytes in the ID portion

I would like to update the detection pattern to handle the new pattern, prior to making the change on NuGet.org

### DIFF
--- a/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
@@ -1,6 +1,6 @@
 [
   {
-    "Pattern": "(^|[^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890+/-_=])(?P<secret>[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{52}JQQJ9(?:9|D|H)[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890][A-L][abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{16}[A-Za-z][abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{7}(?:[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{2}==)?)",
+    "Pattern": "(^|[^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890])(?P<secret>[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{52}JQQJ9(?:9|D|H)[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890][A-L][abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{16}[A-Za-z][abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{7}(?:[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{2}==)?)",
     "Id": "SEC101/200",
     "Name": "CommonAnnotatedSecurityKey",
     "Signatures": [
@@ -319,7 +319,7 @@
   },
   {
     "Pattern": "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)",
-    "Id": "SEC101/030",
+    "Id": "SEC101/031",
     "Name": "NuGetApiKey",
     "Signatures": [
       "oy2"
@@ -346,7 +346,7 @@
     "RegexNormalizedSignature": "AZEG",
     "KeyLength": 32,
     "EncodeForUrl": false,
-    "Id": "SEC101/190",
+    "Id": "SEC101/199",
     "Name": "AzureEventGridIdentifiableKey",
     "DetectionMetadata": "Identifiable"
   },
--- a/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
@@ -318,11 +318,12 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(?i)(^|[^a-z0-9])oy2[a-z2-7]{43}([^a-z0-9]|$)",
+    "Pattern": "(?i)(^|[^a-z0-9])(?<refine>oy2[a-z2-7]{43})([^a-z0-9]|$)",
     "Id": "SEC101/031",
     "Name": "NuGetApiKey",
     "Signatures": [
-      "oy2"
+      "oy2",
+      "OY2"
     ],
     "DetectionMetadata": "HighEntropy, FixedSignature, HighConfidence"
   },
--- a/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
@@ -318,7 +318,7 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(^|[^0-9a-z])(?<refine>oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4])([^aeimquy4]|$)",
+    "Pattern": "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)",
     "Id": "SEC101/030",
     "Name": "NuGetApiKey",
     "Signatures": [
--- a/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/HighConfidenceSecurityModels.json
@@ -318,7 +318,7 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)",
+    "Pattern": "(?i)(^|[^a-z0-9])oy2[a-z2-7]{43}([^a-z0-9]|$)",
     "Id": "SEC101/031",
     "Name": "NuGetApiKey",
     "Signatures": [
--- a/GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
+++ b/GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
@@ -318,11 +318,12 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(?i)(^|[^a-z0-9])oy2[a-z2-7]{43}([^a-z0-9]|$)",
+    "Pattern": "(?i)(^|[^a-z0-9])(?<refine>oy2[a-z2-7]{43})([^a-z0-9]|$)",
     "Id": "SEC101/031",
     "Name": "NuGetApiKey",
     "Signatures": [
-      "oy2"
+      "oy2",
+      "OY2"
     ],
     "DetectionMetadata": "HighEntropy, FixedSignature, HighConfidence"
   },
--- a/GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
+++ b/GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
@@ -318,7 +318,7 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(^|[^0-9a-z])(?<refine>oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4])([^aeimquy4]|$)",
+    "Pattern": "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)",
     "Id": "SEC101/030",
     "Name": "NuGetApiKey",
     "Signatures": [
--- a/GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
+++ b/GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
@@ -318,7 +318,7 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)",
+    "Pattern": "(?i)(^|[^a-z0-9])oy2[a-z2-7]{43}([^a-z0-9]|$)",
     "Id": "SEC101/031",
     "Name": "NuGetApiKey",
     "Signatures": [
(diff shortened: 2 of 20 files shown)


GeneratedRegexPatterns/HighConfidenceSecurityModels.json
```diff
@@ -318,7 +318,7 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(^|[^0-9a-z])(?<refine>oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4])([^aeimquy4]|$)",
+    "Pattern": "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)",
```
michaelcfanning: >A-Z [](http://example.com/codeflow?start=64&length=3)

I'm not sure I'd jump through this hoop. I think I understand what's on your mind, you want to prevent identifying a sequence of characters in some other format, like base64, hex or whatever. If that's correct, though, shouldn't you consider adding the special chars for both standard and URL safe base64? 

Maybe what you have in mind here is to prevent long sequences of all alphabetic chars, which would be entirely fair, that seems like the most likely source of a false positive (it's mathematically uncertain a truly randomized sequence of encoded chars for which n upper-case letter was feasible didn't have at least one for a 43 char sequence). But the `oy2` prefix here is a pretty good preventive against that possibility.

I mention all the above because we haven't really established a good set of principles/conventions for specifying prefix and suffix delimiting regexes. Maybe you can help with that topic, any thoughts?

michaelcfanning: >2-7 [](http://example.com/codeflow?start=49&length=3)

What? So our original 'too clever' regex didn't properly reflect a 32 character encoding. Nuts. #Closed

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -842,7 +842,9 @@ public void SecretMasker_PotentialSymmetricKeysAreClassified(string input, strin
     [DataTestMethod]
     [DataRow("ddddddddddddddddddddddddddddddddddddddddddddAzFu182vhA==", "SEC101/158.AzureFunctionIdentifiableKey:FF8E9A7C2A792029814C755C6704D9427F302E954DEF0FD5EE649BF9163E1F24")]
     [DataRow("eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee+ACRCTB7t/", "SEC101/176.AzureContainerRegistryIdentifiableKey:CE62C55A2D3C220DA0CBFE292B5A6839EC7F747C5B5A7A55A4E5D7D76F1C7D32")]
-    [DataRow("oy2mdeaddeaddeadeadqdeaddeadxxxezodeaddeadwxuq", "SEC101/031.NuGetApiKey:FC93CD537067C7F452073F24C7043D5F58E11B6F49546316BBE06BAA5747317E")]
+    [DataRow("oy2mdeaddeaddeadeadqdeaddeadxxxezodeaddeadwxuq", "SEC101/031.NuGetApiKey:FC93CD537067C7F452073F24C7043D5F58E11B6F49546316BBE06BAA5747317E")] // Fabricated input
+    [DataRow("oy2efu5r4dskzrpeslkab2aqhnwdyr4efhcnvvfx2qgkhm", "SEC101/031.NuGetApiKey:FC93CD537067C7F452073F24C7043D5F58E11B6F49546316BBE06BAA5747317E")] // V4, non-zero byte ID part, GUID-based password part
```
michaelcfanning: >oy2efu5r4dskzrpeslkab2aqhnwdyr4efhcnvvfx2qgkhm [](http://example.com/codeflow?start=14&length=46)

So generally  we prefer not to encode data in source that precisely maps to generated key format.

A convention is to provider keys with repeated patterns that make it obvious we're not looking at randomized data. 

You can update the `GenerateTruePositiveExamples` method in the rule to produce a few of these.

These generated keys will flow to other automated tests that automatically perform testing, you don't need to add tests here. These tests are legacy hard-coded inlined tests from ADO that were put in place to ensure perfect compatibility with them before we switched to our package (that's actually in flight now).

oy11111111111111111111111111111111111111111
oyqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
joelverhagen: It's all good. In the end, we (nuget.org) handled the scale of both patterns with GitHub Secret scanning so that false positive concessions each pattern has is not a problem for us. I think the part I am ignorant about is how other usages of this library are impacted by a false positive in NuGet API keys.

GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
```diff
@@ -318,7 +318,7 @@
     "DetectionMetadata": "Identifiable"
   },
   {
-    "Pattern": "(^|[^0-9a-z])(?<refine>oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4])([^aeimquy4]|$)",
+    "Pattern": "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)",
```
michaelcfanning: >refine [](http://example.com/codeflow?start=35&length=6)

I would have resolved your code conflicts if this had been a branch off main. We are not so large/formal a repo to require people to contribute from forks. Yet?

Anyway, you are now a writer on this repo if you care to simply branch off main when you resolve file conflicts.

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
michaelcfanning: >Pattern [](http://example.com/codeflow?start=5&length=7)

FYI, historically our team has maintained the GitHub patterns and we could contribute this update there if you don't see any action on their side. I will go see if my GitHub contributor perms are still functional, it's been a while since I made a change there.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
joelverhagen: Got it. I'll make this change.

GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
joelverhagen: Since the PR is open and we have comments now, I will continue with this fork on this one. Having all of the relevant comments in one place is better than closing and re-opening the PR I think.

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
joelverhagen: That's great to know. I had it as an action item on my side to request update on our GitHub pattern after this PR is merged. I have gotten traction with their support email in the past so I anticipate no problem. I think we need to maintain that relationship anyway since there is a live endpoint we have to maintain and keep GitHub in sync with (e.g. correct URL if infra changes).

joelverhagen: This is a good discussion. Generally I am not sure if we can know if one solution is better than another without testing it on a large corpus such as all public GitHub commits and measuring false positive rate.

The intent of `[^a-zA-Z0-9]` group at the beginning and end are to find the 46 character base32 string in a config or code file, bounded by non-word characters. The base64 is probably the most common false positive source if I were to guess. Hex could already not match since `oy2` is not hex. Perhaps the intent would be better here if we used `[^\w]` (non-word characters). 

I tried a simple regex on grep.app:
https://grep.app/search?q=oy2%5Ba-z2-7%5D%7B43%7D&regexp=true (this is case insensitive and not exactly the same regex) but it shows all the examples are base64 data.

If we include the missing base64 characters (`+`, `/`, `-`, `_`) we would potentially reduce our FP rate but it's not clear to me that true positives wouldn't have, say, a leading hyphen character... I think I want to keep it as-is just to catch more potential true positives. LMK if I am misunderstanding you.

I am uncertain on how much thought we should put into reducing false positives. To your point, perhaps the `oy2` prefix is already very strong and gets us "99% the way there". 

> I mention all the above because we haven't really established a good set of principles/conventions for specifying prefix and suffix delimiting regexes. Maybe you can help with that topic, any thoughts?

I think my team would have benefitted greatly by these principles/conventions if they existed. I think the only useful thoughts I have on it are:
1. Having an identifiable substring for your secret is a must, unless you want to handle a huge volume of FPs. FP for GitHub Secret scanning is a matter of scaling your integration endpoint, but for client-side tooling it is a real problem. You don't want your users turning off client side secret scanning due to too many FPs.
2. This is more of a question than an idea -- but what do you think about having multiple patterns? Perhaps one with a high FP rate for GitHub Secret Scanning and another with a low FP rate for local tooling. I'm imagining IDE integration where you get a build warning if a secret is found in your code. This would put the high FP rate on the service which can handle it with automation without burning out users on client side tooling with high FP.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
joelverhagen: I removed these test changes and enhanced `GenerateTruePositiveExamples`.

GeneratedRegexPatterns/PreciselyClassifiedSecurityKeys.json
joelverhagen: Fixed merge conflicts.

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
michaelcfanning: Ah, so GitHub forwards your exposures directly to you, I see that in source. Do you auto-revoke on receipt of findings? Do you have any data on scale of reporting to you?

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
```diff
@@ -1,24 +1,37 @@
...
         {
             Id = "SEC101/031";
             Name = nameof(NuGetApiKey);
-            DetectionMetadata = DetectionMetadata.FixedSignature | DetectionMetadata.HighEntropy| DetectionMetadata.HighConfidence;
-            Pattern = "(^|[^0-9a-z])(?<refine>oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4])([^aeimquy4]|$)";
+            DetectionMetadata = DetectionMetadata.FixedSignature | DetectionMetadata.HighEntropy | DetectionMetadata.HighConfidence;
+
+            // This is the ApiKeyV4 format implemented here:
+            // https://github.com/NuGet/NuGetGallery/blob/main/src/NuGetGallery.Services/Authentication/ApiKeyV4.cs
+            Pattern = "(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)";
```
michaelcfanning: >(^|[^a-zA-Z0-9])(?<refine>oy2[a-z2-7]{43})([^a-zA-Z0-9]|$)" [](http://example.com/codeflow?start=23&length=59)

[https://github.com/search?q=%2F%28%3F-i%29%28%5E%7C%5B%5Ea-z0-9%2B%5C%2F%5C-_%5D%29oy2%5Ba-z2-7%5D%7B43%7D%28%24%7C%5B%5Ea-z0-9%2B%5C%2F%5C-_%5D%29%2F&type=code](https://github.com/search?q=%2F%28%3F-i%29%28%5E%7C%5B%5Ea-z0-9%2B%5C%2F%5C-_%5D%29oy2%5Ba-z2-7%5D%7B43%7D%28%24%7C%5B%5Ea-z0-9%2B%5C%2F%5C-_%5D%29%2F&type=code)

Playing around with this in GH, I think we should go ahead and add all base64 encoding relevant chars, i.e. Let me know what you think.

(?-i)(^|[^a-z0-9+\/\-_])oy2[a-z2-7]{43}([^a-z0-9+\/\-_]|$)

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
joelverhagen: The scale is relatively low. We have on the order of 100 notifications per week. I don't have data on the FP rate.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
joelverhagen: This looks reasonable to me! I'll add it to the PR.

michaelcfanning: This does not account for you concern that a user might have a leading or trailing hyphen separating the token from other data. Just want to acknowledge that. I just didn't feel like it made sense to exclude only this character. It 'feels' better to add all four special base64 characters. Alternately, we could simply include `+` and `/` (and not have special delimiter logic for the hyphen and underscore. I suspect this latter approach would be just fine and may in fact be the safest change. I leave it to you, your thoughts?

joelverhagen: In my opinion, if there is any lack of clarity at this stage, we should take the reasonable option that has the lowest chance of missing a true positive. Along those lines, a NuGet API key bounded by alphanumeric characters seems reasonable to miss because I can't think of any config file or script I've used in the past where a string bound is `[a-z0-9]`, whereas `+`, `/`, `-`, or `_` all feel like visual boundaries on a string different from alphanumeric (although definitely less common than say a whitespace character or a quote character).

I think your mention of case insensitivity is important since, upon inspection, it looks like our code is actually case _insensitive_ on API keys (!!!). Meaning `OY2...` is accepted by our endpoint. I just tested it. We do hand out lowercase base32 when a user creates an API key but if for some crazy reason if they uppercase their API key in their config file/script, it would still work. https://github.com/NuGet/NuGetGallery/blob/7c81548db51ed543ff8f6558c1306571a6bb368a/src/NuGetGallery.Services/Authentication/ApiKeyV4.cs#L124-L126

So if it is left to me, I would do this:
```
(?i)(^|[^a-z0-9])oy2[a-z2-7]{43}([^a-z0-9]|$)
```

image
https://regex101.com/r/GALkvJ/1

Thus we are accepting that some unlucky base64 will be considered as a NuGet API key and will get flagged (and sent to our secret invalidation endpoint). If we see this cause problems in the future in any usage of this library or with GitHub Secret Scanning, we can reassess this decision with that clear case in mind.

joelverhagen: As a meta point, I believe this decision is a not a very scary one (between any of the regexes we have come up with) because we are wanting to move to HISv2 anyway :)

michaelcfanning: Alright, sorry to rathole on this. This library is used for redaction purposes, and so it is actually a concern if we remove unlucky base64 data. We do see evidence of noisy hits in the wild, the question is whether we will comfortable redacting that data from, say, a build log or telemetry stream. 

For sure, GitHub should receive a looser version, in that context, no data is removed, any finding can be bypassed, and any reported finding can be easily filtered by the handler (you, NuGet).

So, this is getting a bit detailed, but one thing we could consider doing is this: we proceed with the pattern that you've suggested, which may find an unlucky base64 pattern embedded in other data. We have a minor bit of post-processing that examines our find and in the case that we determine the casing of the token is inconsistent (i.e., we see a mix of upper and lower case), we reject the find as unlucky base64 data.

Now, if your back-end is completely case-insensitive, a user could mix casing arbitrarily. But it seems like it could be a rare occurrence someone would partly modify a token in this way. 

It is non-obvious but `GetMatchIdAndName` allows for this post-processing. This helper receives a match. If you reject it, return null, otherwise let the base class handle things (it will simply construct moniker for the find, which is the combined rule id + friendly name).

See this code for an example:

[https://github.com/microsoft/security-utilities/blob/034040c8038c27dbac3bbe1fd8e2bcec37f0ffa2/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_102_AdoPat.cs#L22](https://github.com/microsoft/security-utilities/blob/034040c8038c27dbac3bbe1fd8e2bcec37f0ffa2/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_102_AdoPat.cs#L22)

joelverhagen: Ratholing on things like this is an indulgence of mine. Thank you for the context, this helps a lot! I see now that there are multiple distinct use cases of secret detection and the pattern accordingly may vary. I will look at the `GetMatchIdAndName` method and try to enhance it for NuGet.

joelverhagen: Pushed another change. I implemented `GenerateFalsePositiveExamples` to get test coverage.

michaelcfanning (approved):
:shipit:

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
michaelcfanning: Beautiful. Ship it. I am frequently asked why there isn't a single converged library of regexes to serve all scenarios but there are some pernicious subtleties in their application that makes it hard to organize patterns to serve all needs. This discussion is a good example.

PR #103, opened 2024-10-29
CASK detection with prefix or suffix

suvamM:
CASK keys with a prefix (such as `https://`) or suffix (such as `@github.com`) would fail detection. This PR fixes that, and adds a test for the same.


nguerrera (dismissed):

nguerrera:
There is a test failure. Is it related?

nguerrera (approved):

PR #109, opened 2024-11-18
Removing base62 rust crate

suvamM:
The `base62` Rust crate depended on the `failure` Rust crate, which has a [known vulnerability](https://github.com/advisories/GHSA-jq66-xh47-j9f3).

### DIFF
--- a/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
+++ b/src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
@@ -5,7 +5,6 @@
 #![allow(dead_code)]
 #![allow(unused_assignments)]
 
-use base_62;
 use base64::{engine::general_purpose, Engine as _};
 use chrono::Datelike;
 use core::panic;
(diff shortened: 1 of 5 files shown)


nguerrera:
Does this need a release note?

nguerrera (dismissed):

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
```diff
@@ -238,7 +237,7 @@ pub fn generate_common_annotated_test_key(
             let mut rng = rand::thread_rng();
             rng.try_fill_bytes(&mut key_bytes).expect("Failed to generate random bytes.");
 
-            key  = base_62::encode(&key_bytes);
+            key  = general_purpose::STANDARD.encode(&key_bytes);
```
nguerrera, line 240: Is there a product impact here?

suvamM:
> Does this need a release note?

@nguerrera Good point. Just added the release notes, and updated the version.

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_secrets.rs
suvamM: The tests still pass, so I don't think there is a regression. Let me cross check what happens in the C# world.

suvamM: @nguerrera There is a check right at the end of the Rust method which checks if there are special characters (+ or /) in the generated key. If yes, the process is repeated. So there is no product impact. Though I see the C# and Rust versions of this function has drifted.

nguerrera (approved):

PR #111, opened 2024-12-04
Allow allocation time to be specified

joelverhagen:
Also fix highly improbably timing bug where allocation year and month do not match.

This will allow NuGet to encode more specific allocation details that are consistent with the built-in year and month values.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -373,7 +373,7 @@ public static class IdentifiableSecrets
         // An allocation time before 2024 or after 2085 is outside of the possible range a base62 character can express.
         if (allocationTime.Year < 2024 || allocationTime.Year > 2085)
         {
-            throw new ArgumentOutOfRangeException(nameof(allocationTime), "The allocation time year must between 2024 and 2085, inclusive.");
+            throw new ArgumentOutOfRangeException(nameof(allocationTime), "The allocation year must be between 2024 and 2085, inclusive.");
         }
 
         base64EncodedSignature = customerManagedKey
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -369,9 +369,11 @@ public static class IdentifiableSecrets
             throw new ArgumentException(nameof(allocationTime), "The allocation time must be in UTC.");
         }
 
-        if (allocationTime.Year < 2024)
+        // 2085 is 61 years after 2024. A base62 character is used to express the year (A = 2024 up to 9 = 2085).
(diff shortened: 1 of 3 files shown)


src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -341,6 +364,16 @@ public static string GenerateCommonAnnotatedTestKey(byte[] randomBytes,
             providerReserved = new byte[providerReservedLength];
         }
 
+        if (allocationTime.Kind != DateTimeKind.Utc)
+        {
+            throw new ArgumentException(nameof(allocationTime), "The allocation time must be in UTC.");
+        }
+
+        if (allocationTime.Year < 2024)
```
michaelcfanning: >allocationTime [](http://example.com/codeflow?start=12&length=14)

We could also add the upper bounds check, i.e., if allocationTime.Year > 87 #Resolved

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -341,6 +364,16 @@ public static string GenerateCommonAnnotatedTestKey(byte[] randomBytes,
             providerReserved = new byte[providerReservedLength];
         }
 
+        if (allocationTime.Kind != DateTimeKind.Utc)
```
michaelcfanning, line 367: We need a release note for this bug. Something along the lines of this, in your own language. See docs\ReleaseHistory.md for the content that needs updating.

BUG: Cache current date to prevent low-likelihood timing issue where year or month changes between two `DateTime.UtcNow` retrievals leads to the wrong encoded timestamp.

michaelcfanning (dismissed):
:shipit:

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
joelverhagen: Good idea, I will add this!

joelverhagen: I think the max allocation year would be 2085, right? Since it's 2024 + 61 (0 - 61 / 'A' - '9' being supported base62 digits).

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
joelverhagen: Added in my own language :)

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
joelverhagen: Fixed.

michaelcfanning: right!

michaelcfanning (approved):
:shipit:

PR #114, opened 2024-12-17
Edits for vso integration

michaelcfanning:
- BUG: Mark `SecretMasker(SecretMasker)` copy contructor as protected to make it callable by derived classes.
- FPS: Update `SEC101/127.UrlCredentials` regex to not fire on use of colon within URL path component.
- BUG: Mark `SecretMasker.LiteralEncoders`, `SecretMasker.EncodedSecretLiterals` and `SecretMasker.ExplicitlyAddedSecretLiterals` as public

### DIFF
--- a/GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
@@ -20,7 +20,7 @@
     "DetectionMetadata": "HighEntropy, MediumConfidence"
   },
   {
-    "Pattern": "(ftps?|https?):\\/\\/(?:[^:@]+):(?<refine>[^:@?]+)@",
+    "Pattern": "(ftps?|https?):\\/\\/(?<refine>[^:@\\/]+:[^:@?\\/]+)@",
     "Id": "SEC101/127",
     "Name": "UrlCredentials",
     "Signatures": [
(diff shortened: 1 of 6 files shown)


rwoll (approved):

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
```diff
@@ -20,7 +20,7 @@
     "DetectionMetadata": "HighEntropy, MediumConfidence"
   },
   {
-    "Pattern": "(ftps?|https?):\\/\\/(?:[^:@]+):(?<refine>[^:@?]+)@",
+    "Pattern": "(ftps?|https?):\\/\\/(?<refine>[^:@\\/]+:[^:@?\\/]+)@",
```
rwoll, line 23: FYI, this is technically a valid URL: `https://httpbin.org/anything/a:b:c` and will route to the path `/anything/a:b:c`. See response when you visit that URL.

rwoll: See also: `https://httpbin.org/anything/https://foo:bar@example.com` which will also match and route. (Though, these are enough of an edge case where that's acceptable and likely desired.)

nguerrera (approved):

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
michaelcfanning: Oof. Yes. It is just this sort of nuance and complexity that makes it hard to develop patterns that work for all scenarios. :) I think we'll go with this and perhaps revisit in the future.

PR #116, opened 2025-01-02
Expose Value property on SecretLiteral

michaelcfanning:
# 1.11.0 - 01/02/2025
- NEW: Expose `SecretLiteral.Value` as public data.
- FPS: Update `SEC101/127.UrlCredentials` regex to require a word break before the `ftp` or `http` schema rendering.

### DIFF
--- a/GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
+++ b/GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
@@ -20,7 +20,7 @@
     "DetectionMetadata": "HighEntropy, MediumConfidence"
   },
   {
-    "Pattern": "(ftps?|https?):\\/\\/(?<refine>[^:@\\/]+:[^:@?\\/]+)@",
+    "Pattern": "($|\\b)(ftps?|https?):\\/\\/(?<refine>[^:@\\/]+:[^:@?\\/]+)@",
     "Id": "SEC101/127",
     "Name": "UrlCredentials",
     "Signatures": [
--- a/src/Microsoft.Security.Utilities.Core/UrlCredentials.cs
+++ b/src/Microsoft.Security.Utilities.Core/UrlCredentials.cs
@@ -14,7 +14,7 @@ public sealed class UrlCredentials : RegexPattern
 
         Name = nameof(UrlCredentials);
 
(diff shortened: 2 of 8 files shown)


GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
```diff
@@ -20,7 +20,7 @@
     "DetectionMetadata": "HighEntropy, MediumConfidence"
   },
   {
-    "Pattern": "(ftps?|https?):\\/\\/(?<refine>[^:@\\/]+:[^:@?\\/]+)@",
+    "Pattern": "($|\\b)(ftps?|https?):\\/\\/(?<refine>[^:@\\/]+:[^:@?\\/]+)@",
```
suvamM, line 23: How are we testing this change to the regex?

suvamM (commented):
There is a test failure. The version needs to be updated in the `SecretMasker_Version` test.

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
michaelcfanning: Did you see the test updates? Let me ensure I pushed them.

src/Microsoft.Security.Utilities.Core/UrlCredentials.cs
```diff
@@ -50,6 +50,8 @@ public override IEnumerable<string> GenerateFalsePositiveExamples()
         {
             $"http://example.com/embedded:colon",
             $"ftp://@example.com/embedded:colon",
+            $"prefixedftp://@example.com/embedded:colon",
```
michaelcfanning, line 53: >prefixedftp [](http://example.com/codeflow?start=14&length=11)

These are the test changes.

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
michaelcfanning: Yes, the tests are further in this file. We have a pattern in this code of generating test data in the rule itself.

suvamM (approved):

cfaucon (approved):

PR #119, opened 2025-01-16
Fix bugs related to erroneous or incorrectly interpreted suffix and prefixes.

michaelcfanning:
- FNS: Eliminate false negatives resulting from incorrectly specifying `=` as a delimiting character in the core 'identifiable' rules. This broke simple patterns such as `myKey=an_actual_key`.
- FNS: Eliminate false negatives resulting from improper use of the `-` character in regexes (where it was interpreted as a range operator not a literal).,

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/IdentifiableKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/IdentifiableKey.cs
@@ -53,14 +53,36 @@ namespace Microsoft.Security.Utilities
                     string encoded = new string(alphabet[i], encodedLength);
                     Array.Copy(Convert.FromBase64String(encoded), bytes, KeyLength);
 
-                    yield return
+                    string key =
                         IdentifiableSecrets.GenerateBase64KeyHelper(checksumSeed,
                                                                     keyLengthInBytes: KeyLength,
                                                                     Signatures!.First(),
                                                                     EncodeForUrl,
                                                                     bytes);
+
+                    yield return key;
+
+                    foreach (string prefix in s_nonInvalidatingPrefixes)
+                    {
+                        yield return $"{prefix}{key}";
+                    }
+
+                    foreach (string suffix in s_nonInvalidatingSuffixes)
+                    {
+                        yield return $"{key}{suffix}";
+                    }
                 }
             }
         }
+
+        private static readonly string[] s_nonInvalidatingPrefixes = new[]
+        {
+            "=",
+        };
+
+        private static readonly string[] s_nonInvalidatingSuffixes = new[]
+        {
+            ";",
+        };
     }
 }
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Security.Utilities
             // specifically, it looks for a key embedded in a connection string. We have had
             // misses previously due to mismanaging delimiters around the identifiable key
             // rules, so we will obtain the legacy rule patterns, replace their secret with
-            // and identifiable one, and ensure this rule continues to function against a very
+            // an identifiable one, and ensure this rule continues to function against a very
             // typical expression of this credential kind.
             foreach (string example in legacyStorageAccountKey.GenerateTruePositiveExamples())
             {
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Security.Utilities
 {
@@ -16,5 +17,35 @@ namespace Microsoft.Security.Utilities
         public override ISet<string> Signatures => IdentifiableMetadata.AzureStorageSignature.ToSet();
 
         public override IEnumerable<ulong> ChecksumSeeds => new[] { IdentifiableMetadata.AzureStorageAccountChecksumSeed };
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            string sampleKey = string.Empty;
+
+            foreach (string example in base.GenerateTruePositiveExamples())
+            {
+                sampleKey = example;
+                yield return example;
+            }
+
+            var legacyStorageAccountKey = new AzureStorageAccountLegacyCredentials();
+
+            // The legacy storage account key rule requires more context in the scan contents,
+            // specifically, it looks for a key embedded in a connection string. We have had
+            // misses previously due to mismanaging delimiters around the identifiable key
+            // rules, so we will obtain the legacy rule patterns, replace their secret with
+            // and identifiable one, and ensure this rule continues to function against a very
+            // typical expression of this credential kind.
+            foreach (string example in legacyStorageAccountKey.GenerateTruePositiveExamples())
+            {
+                string standaloneSecret =
+                        CachedDotNetRegex.Instance.Matches(example,
+                                                           legacyStorageAccountKey.Pattern,
+                                                           captureGroup: "refine").First().Value;
+
+                string generatedExample = example.Replace(standaloneSecret, sampleKey);
+                yield return generatedExample;
+            }
+        }
     }
 }
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
@@ -42,7 +42,6 @@ namespace Microsoft.Security.Utilities
 
                     yield return key;
 
-
                     foreach (string prefix in s_nonInvalidatingPrefixes)
                     {
                         yield return $"{prefix}{key}";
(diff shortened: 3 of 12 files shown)


src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
```diff
@@ -16,5 +17,35 @@ public AzureStorageAccountIdentifiableKey()
...
+                yield return example;
+            }
+
+            var legacyStorageAccountKey = new AzureStorageAccountLegacyCredentials();
+
+            // The legacy storage account key rule requires more context in the scan contents,
+            // specifically, it looks for a key embedded in a connection string. We have had
+            // misses previously due to mismanaging delimiters around the identifiable key
+            // rules, so we will obtain the legacy rule patterns, replace their secret with
+            // and identifiable one, and ensure this rule continues to function against a very
```
nguerrera:
```suggestion
            // an identifiable one, and ensure this rule continues to function against a very
```

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -158,7 +158,7 @@ public static string GenerateString(string alphabet, int count)
     private const string SparseAlpha = $"{SparseLowercase}{sparseUppercase}";
     private const string SparseBase62 = $"{SparseAlpha}{SparseDigit}";
     private const string SparseBase64 = $"{SparseBase62}+"; // Forward slash elided.
-    private const string SparseUrlSafeBase64 = $"{SparseBase62}-"; // Underscore elided.
+    private const string SparseUrlSafeBase64 = @$"{SparseBase62}\-"; // Underscore elided.
```
michaelcfanning, line 164: >\- [](http://example.com/codeflow?start=64&length=2)

Hyphens in literals that end up in a regex range maybe interpreted in unexpected ways. A hyphen at the end of the sequence will consistent be interpreted as a literal. A hyphen within a character sequence might end up being interpreted as a range operator. In this change we both explicitly escape all hyphens and then also move them to the end of the literal.

This double defensiveness might not be necessary. Previously, the hyphens were placed where they were to indicate their ordering in formats such as base64. In reality, thought, these sequences aren't consistent between base62 and base64 in any case, so there's no substantive value in trying to communicate anything with ordering.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
```diff
@@ -25,24 +25,45 @@ public override IEnumerable<string> GenerateTruePositiveExamples()
...
+                                                                                 providerReserved: null,
+                                                                                 longForm,
+                                                                                 testChar,
+                                                                                 keyKindSignature);
                     }
 
-                    yield return example;
+                    yield return key;
+
+
```
nguerrera: nit: extra newline here

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
```diff
@@ -16,5 +17,35 @@ public AzureStorageAccountIdentifiableKey()
...
+
+            foreach (string example in base.GenerateTruePositiveExamples())
+            {
+                sampleKey = example;
+                yield return example;
+            }
+
+            var legacyStorageAccountKey = new AzureStorageAccountLegacyCredentials();
+
+            // The legacy storage account key rule requires more context in the scan contents,
```
michaelcfanning, line 33: >legacy [](http://example.com/codeflow?start=19&length=6)

This comment reflects a prominent problem report. An azure storage connection string expresses the key as an unquoted token following the pattern `AccountKey=`. Because we improperly included the equals sign as a forbidden prefixed delimiter, we dropped these finds.

AccountKey=SOMEKEY

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -183,8 +183,8 @@ public static string GenerateString(string alphabet, int count)
     public const string SuffixUrlUnreserved = $"([^{RegexEncodedUrlUnreserved}+/=]|{End})";
     public const string PrefixBase62 = $"({Start}|[^{Base62}])";
     public const string SuffixBase62 = $"([^{Base62}]|{End})";
-    public const string PrefixAllBase64 = $"({Start}|[^{Base64}-_=])";
```
michaelcfanning, line 186: >= [](http://example.com/codeflow?start=65&length=1)

Second source of bugs, we improperly forbade the equal sign as a prefix character, missing the obvious `something=secret` case (common in connection strings in storage, cosmosdb, etc.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/IdentifiableKey.cs
```diff
@@ -53,14 +53,36 @@ public override IEnumerable<string> GenerateTruePositiveExamples()
...
+
+                    foreach (string suffix in s_nonInvalidatingSuffixes)
+                    {
+                        yield return $"{key}{suffix}";
+                    }
                 }
             }
         }
+
+        private static readonly string[] s_nonInvalidatingPrefixes = new[]
```
michaelcfanning, line 78: >s_nonInvalidatingPrefixes [](http://example.com/codeflow?start=41&length=25)

i've included the characters that were observed in the wild to provoke an issue. presumably, fixing these fixes the full range of improperly performing chars in play, since we no longer have hyphen literal vs. range operator confusion.

and so, I didn't bother to invest anything in trying to produce a comprehensive set of things to test here.

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -792,14 +792,19 @@ public void IdentifiableSecrets_DerivedSymmetricKeys()
                     {
                         foreach (string signature in identifiablePattern.Signatures)
                         {
-                            if (IdentifiableSecrets.TryValidateBase64Key(securityKey, checksumSeed, signature, identifiablePattern.EncodeForUrl))
+                            string standaloneSecret =
```
michaelcfanning, line 826: >standaloneSecret [](http://example.com/codeflow?start=35&length=16)

When we produce an example, it sometimes but not always comprises a standalone key. Here is a simple pattern for obtain the secret, we grab the output of the <refine> group.

I put a second pattern for doing this elsewhere just to compare.

nguerrera (dismissed):

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -175,13 +175,18 @@ private void ValidateSecurityModelsMasking(IEnumerable<RegexPattern> patterns, I
 
                         foreach (string testExample in pattern.GenerateTruePositiveExamples())
                         {
-                            string secretValue = testExample;
+                            var detection = secretMasker.DetectSecrets(testExample).FirstOrDefault();
```
michaelcfanning, line 177: >detection [](http://example.com/codeflow?start=32&length=9)

Here's a second way to obtain the literal secret. Run the detection and use the location data to get the secret value. 

This approach is much more what you'd expect in our API scenarios. I've left this inconsistent consciously. The other approach is more concise and expedient for tests.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
```diff
@@ -16,5 +17,35 @@ public AzureStorageAccountIdentifiableKey()
         public override ISet<string> Signatures => IdentifiableMetadata.AzureStorageSignature.ToSet();
 
         public override IEnumerable<ulong> ChecksumSeeds => new[] { IdentifiableMetadata.AzureStorageAccountChecksumSeed };
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
```
rwoll, line 21: Your release note has the example of `myKey=an_actual_key`, but I don't see that literal example in a test.

While your code looks like it generates something similar, please include literal examples in addition to the generated ones.

michaelcfanning: This won't actually trigger anything, as is probably obvious. You need an actual key to trigger and an actual key will get blocked and read as bad signal in the code.  This code base produces very high fidelity test patterns on demand by calling the literal API used by providers to generate keys. As such, we have very little hard-coded test examples (to keep that stuff out of source). The fictional provider could be used to do more with hard-coded tests. And we could do more to produce test keys for existing providers using the same 'trick' of populating all random data with a single character. We should build an API for this and also harden our scan reports to filter away this apparently valid but benign data. 

I didn't create one-offs for this pattern because if I create a special delimiter test for one 64-byte HIS key, why not all? and so I updated the low level generator to opt into this testing for all those keys (32-byte as well). This is powerful but has the disadvantage of not making certain test conditions explicit (and therefore readable). I acknowledge this as an area we should consider improving, for sure, let's talk more.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
michaelcfanning: gotcha

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_152_AzureStorageAccountIdentifiableKey.cs
nguerrera: I think it's definitely worth investing here. There is a lot of value in tests with fixed data on top of test data generation.

nguerrera (dismissed):

cfaucon (dismissed):

cfaucon (approved):

PR #122, opened 2025-01-29
Run tests on .NET 8, fix issues uncovered by that

nguerrera:
This started with wanting to examine `IdentifiableScan` performance. I profiled its unit test and the profiler told me that it was spending a lot of time in regular expressions and so I probably wanted to use a newer version of .NET. :) This made me realize that the tests were running only on .NET Framework. I enabled them on .NET 8 and when I ran them, I was greeted with `System.NotSupportedException: The specified pattern with RegexOptions.NonBacktracking could result in an automata as large as 'NNNN' nodes, which is larger than the configured limit of '1000'.`

This was due to `ValidateBase64Key` generating a regex with a large quantifier for large input. I fixed this by removing the regex generation and use. Instead, we only check that 1) there is not inconsistent use of base64url or standard base64 with respect to the `encodeForUrl` argument, and 2) the key has the signature provided. The remainder of the work done by the generated regex was redundant to what had already been ascertained by `ValidateChecksum`. In addition to fixing the bug, this improved the performance of the slowest test, `IdentifiableSecrets_GenerateBase64Key_Comprehensive`, by about 8x on .NET Framework.

While looking at this, I further discovered that `ValidateBase64Key`  did not check the signature if given a backwards-compatible `CommonAnnotatedKey`, and that `ValidateCommonAnnotatedKey` didn't check the signature at all. Also, the `byte[]` version of ValidateCommonAnnotatedKey did not validate the signature argument. These bugs are fixed here as well.

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -14,7 +14,7 @@
 # 1.13.0 - UNRELEASED
 - BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey(byte[], string)` did not validate signature argument to be be exactly 4 characters long, beginning with a letter, entirely alphanumeric, and either entirely uppercase or entirely lowercase. 
 - BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey` (all overloads)  did not check that the key had the given signature and would return true for any valid key.
-- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key`, when given a backwards-compatible `CommonAnnotatedKey`, also did not check that the key had the given signature.
+- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key`, when given a backwards-compatible `CommonAnnotatedKey`, did not check that the key had the given signature.
 - BUG: `IdentifiableSecrets.(Try)ValidateBase64Key` failed on .NET 8 with `System.NotSupportedException: The specified pattern with RegexOptions.NonBacktracking could result in an automata as large as 'NNNN' nodes, which is larger than the configured limit of '1000'.`
 - PRF: `IdentifiableSecrets.(Try)ValidateBase64Key` is much faster now as it no longer generates nor uses regular expressions.
 
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -15,7 +15,7 @@
 - BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey(byte[], string)` did not validate signature argument to be be exactly 4 characters long, beginning with a letter, entirely alphanumeric, and either entirely uppercase or entirely lowercase. 
 - BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey` (all overloads)  did not check that the key had the given signature and would return true for any valid key.
 - BUG: `IdentifiableSecrets.(Try)ValidateBase64Key`, when given a backwards-compatible `CommonAnnotatedKey`, did not check that the key had the given signature.
-- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key` failed on .NET 8 with `System.NotSupportedException: The specified pattern with RegexOptions.NonBacktracking could result in an automata as large as 'NNNN' nodes, which is larger than the configured limit of '1000'.`
+- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key` failed on .NET 8 with `System.NotSupportedException: The specified pattern with RegexOptions.NonBacktracking could result in an automata as large as 'NNNN' nodes, which is larger than the configured limit of '1000'.` This was due to using a regex with a quantifier that was too large, and fixed by removing the regex entirely.
 - PRF: `IdentifiableSecrets.(Try)ValidateBase64Key` is much faster now as it no longer generates nor uses regular expressions.
 
 # 1.12.0 - 01/06/2025
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -11,6 +11,13 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# 1.13.0 - UNRELEASED
+- BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey(byte[], string)` did not validate signature argument to be be exactly 4 characters long, beginning with a letter, entirely alphanumeric, and either entirely uppercase or entirely lowercase. 
+- BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey` (all overloads)  did not check that the key had the given signature and would return true for any valid key.
+- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key`, when given a backwards-compatible `CommonAnnotatedKey`, also did not check that the key had the given signature.
+- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key` failed on .NET 8 with `System.NotSupportedException: The specified pattern with RegexOptions.NonBacktracking could result in an automata as large as 'NNNN' nodes, which is larger than the configured limit of '1000'.`
+- PRF: `IdentifiableSecrets.(Try)ValidateBase64Key` is much faster now as it no longer generates nor uses regular expressions.
+
 # 1.12.0 - 01/06/2025
 - BRK: Derived keys and hashed data are no longer supported. The following API are removed:
   - `IdentifiableSecrets.CommonAnnotatedDerivedKeySignature`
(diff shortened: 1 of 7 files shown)


rwoll (commented):
LGTM, but please have Mike review as well.

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,13 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# 1.13.0 - UNRELEASED
+- BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey(byte[], string)` did not validate signature argument to be be exactly 4 characters long, beginning with a letter, entirely alphanumeric, and either entirely uppercase or entirely lowercase. 
+- BUG: `IdentifiableSecrets.TryValidateCommonAnnotatedKey` (all overloads)  did not check that the key had the given signature and would return true for any valid key.
+- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key`, when given a backwards-compatible `CommonAnnotatedKey`, also did not check that the key had the given signature.
```
michaelcfanning: >also [](http://example.com/codeflow?start=109&length=4)

NIT: remove use of `also` here

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,13 @@
...
+- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key` failed on .NET 8 with `System.NotSupportedException: The specified pattern with RegexOptions.NonBacktracking could result in an automata as large as 'NNNN' nodes, which is larger than the configured limit of '1000'.`
```
michaelcfanning: >NotSupportedException [](http://example.com/codeflow?start=82&length=21)

We don't have to use this convention but our other scanners provide specific change codes for unhandled exceptions, differentiating between those raised by a specific rule (which typically disables that rule) or an engine-level exception.

For the current library (and change) we have general API as well as a classifier/scan engine.

A reason for categorizing exceptions in general is simply that these symptoms are typical/often searched for in release notes. What's most important is to author a good/consistent release note. The current one is good! Except that I might have added a bit more about why the exception occurred and why it no longer does. E.g. that the base64 regex had too large a quantifier and that the situation is resolved ultimately because the regex use was entirely removed.



- UER => Eliminate unhandled exceptions in rule.
- UEE => Eliminate unhandled exceptions in engine. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -334,6 +336,11 @@ public void IdentifiableSecrets_TryValidateCommonAnnotatedKeyBase64EncodedChecks
                     bool result = IdentifiableSecrets.TryValidateCommonAnnotatedKey(keyBytes, signature);
                     result.Should().BeTrue(because: $"{key}' should validate using 'TryValidateCommonAnnotatedKey' with its original checksum");
 
+                    string differentSignature = GetRandomSignature();
+                    differentSignature.Should().NotBe(signature, because: "it is random.");
```
michaelcfanning: >. [](http://example.com/codeflow?start=87&length=1)

NIT: trailing periods typically don't render well in these API, so we don't add them.

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -1004,6 +1021,29 @@ public void IdentifiableSecrets_GenerateBase64Key_ShouldThrowExceptionForInvalid
...
+            result.Should().Be(true, because: "validation of standard base64 key with encodeForUrl=false should succeed");
+
+            result = IdentifiableSecrets.TryValidateBase64Key(secretBase64Url, seed, signature, encodeForUrl: true);
+            result.Should().Be(true, because: "validation of base64url key with encodeForUrl=true should succeed");
+
+            result = IdentifiableSecrets.TryValidateBase64Key(secretBase64Std, seed, signature, encodeForUrl: true);
+            result.Should().Be(false, because: "validation of standard base64 key with encodeForUrl=true should fail");
+
+            result = IdentifiableSecrets.TryValidateBase64Key(secretBase64Url, seed, signature, encodeForUrl: false);
+            result.Should().Be(false, because: "validation of base64url key with econdeForUrl=false should fail");
```
michaelcfanning: >econdeForUrl [](http://example.com/codeflow?start=81&length=12)

`encode`

michaelcfanning (dismissed):
:shipit:

docs/ReleaseHistory.md
```suggestion
- BUG: `IdentifiableSecrets.(Try)ValidateBase64Key`, when given a backwards-compatible `CommonAnnotatedKey`, did not check that the key had the given signature.
```

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
nguerrera:
```suggestion
                    differentSignature.Should().NotBe(signature, because: "it is random");
```

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -334,6 +336,11 @@ public void IdentifiableSecrets_TryValidateCommonAnnotatedKeyBase64EncodedChecks
                     bool result = IdentifiableSecrets.TryValidateCommonAnnotatedKey(keyBytes, signature);
                     result.Should().BeTrue(because: $"{key}' should validate using 'TryValidateCommonAnnotatedKey' with its original checksum");
 
+                    string differentSignature = GetRandomSignature();
+                    differentSignature.Should().NotBe(signature, because: "it is random.");
+                    result = IdentifiableSecrets.TryValidateCommonAnnotatedKey(keyBytes, differentSignature);
+                    result.Should().BeFalse(because: $"'{key}' has signature '{signature}', not '{differentSignature}'.");
```
nguerrera:
```suggestion
                    result.Should().BeFalse(because: $"'{key}' has signature '{signature}', not '{differentSignature}'");
```

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -461,8 +468,12 @@ public void IdentifiableSecrets_TryValidateCommonAnnotatedKey_GenerateCommonAnno
...
                                                                                  longForm: true);
 
-                bool result = TryValidateCommonAnnotatedKeyHelper(validKey, validSignature, out string failedApi);
-                result.Should().BeTrue(because: $"'{validKey}' should validate using '{failedApi}'");
+                bool result = TryValidateCommonAnnotatedKeyHelper(validKey, validSignature);
+                result.Should().BeTrue(because: $"'{validKey}' should validate.");
+
+                string differentSignature = "WXYZ";
+                result = TryValidateCommonAnnotatedKeyHelper(validKey, differentSignature);
+                result.Should().BeFalse(because: $"'{validKey}' has signature '{validSignature}', not '{differentSignature}'.");
```
nguerrera:
```suggestion
                result.Should().BeFalse(because: $"'{validKey}' has signature '{validSignature}', not '{differentSignature}'");
```

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
nguerrera:
```suggestion
            result.Should().Be(false, because: "validation of base64url key with encodeForUrl=false should fail");
```

docs/ReleaseHistory.md
nguerrera: Should I add both of these to legend and then classify this as `UEE`?

nguerrera: > Except that I might have added a bit more about why the exception occurred and why it no longer does.

Done, leaving only the question above about the categorization before all comments are resolved.

src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -521,9 +535,12 @@ public void IdentifiableSecrets_TryValidateCommonAnnotatedKey_RejectInvalidSigna
                                                                              new byte[9],
                                                                              new byte[3]);
 
-            bool result = IdentifiableSecrets.TryValidateCommonAnnotatedKey(validKey, validSignature);
+            bool result = TryValidateCommonAnnotatedKeyHelper(validKey, validSignature);
             result.Should().BeTrue(because: "a generated key should validate");
 
+            result = TryValidateCommonAnnotatedKeyHelper(validKey, base64EncodedSignature: "aBcD");
+            result.Should().BeFalse(because: "although the signature comparison is case-insentitive, the signature argument must have consistent case.");
```
nguerrera:
```suggestion
            result.Should().BeFalse(because: "although the signature comparison is case-insentitive, the signature argument must have consistent case");
```

michaelcfanning (approved):
:shipit:

docs/ReleaseHistory.md
nguerrera: Per offline discussion, we're leaving this for now.

PR #123, opened 2025-01-30
build: speed up BuildAndTest.cmd

nguerrera:
`BuildAndTest` was terribly slow and basically unusable in any inner loop. The main culprit was that `cargo clean` was done on every invocation. This causes the rust world to be rebuilt every single time, and that is quite slow. There were some other issues that are also fixed here.

Changes:
* Remove all `cargo clean` calls. Either use `git clean` or run `cargo clean` yourself if you want to rebuild. This is consistent with the C# behavior of `BuildAndTest`.
* Stop doing `cargo test` on the `ffi` directory as this builds stuff again unnecessarily but runs no additional tests.
* Extract a separate BuildAndTestRust.cmd script to allow iteration on rust changes without running C# build and test steps.
* Respect `-NoTest` argument for rust.
* Clean up some build error handling and logging output.
* Remove dead `rd /s /q` code. If you had a `<repo>\..\bld`, which you usually don't, it would fail because this is cmd not powershell syntax. Furthermore, blowing away `bld` n every invocation runs counter to the goal here of speeding things up for iteration, so it is better to delete this thing that never did anything than to fix it.

### DIFF
--- a/BuildAndTest.cmd
+++ b/BuildAndTest.cmd
@@ -1,47 +1,18 @@
 @echo off
-time /T
+setlocal
 
-pushd .\src\security_utilities_rust\
-call cargo clean --release
-call cargo test --release
-if "%ERRORLEVEL%" NEQ "0" (
-  echo "security_utilities_rust testing failed..."
-  exit /b %ERRORLEVEL%
-)
-popd
+time /t
+set RESULT=1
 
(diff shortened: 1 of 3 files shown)


nguerrera:
#124 will fix the failed PR validation.

rwoll (approved):

PR #124, opened 2025-01-30
ci: fix build break due to deprecated v3 artifact upload

nguerrera:
See https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

Fixed by using v4.

### DIFF
--- a/.github/workflows/test-marvin-nodejs.yml
+++ b/.github/workflows/test-marvin-nodejs.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Run tests
         run: npm test
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         if: always()
         with:
           name: playwright-report


.github/workflows/test-marvin-nodejs.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       - name: Run tests
         run: npm test
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
```
nguerrera, line 29: @rwoll It seems this doesn't actually upload anything as it doesn't find `playwright-report`. This was already the case with v3 too from what I can tell. Should I remove it instead of upgrading it to v4?

rwoll: If playwright test fails, the trace/failure logs get uploaded via this. So I recommend leaving it.

rwoll (approved):

.github/workflows/test-marvin-nodejs.yml
nguerrera: Ah!

PR #126, opened 2025-02-04
test: use string interpolation in place of hard-coded signature

nguerrera:
This works around an internal false positive scanning issue.

### DIFF
--- a/src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
+++ b/src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
@@ -1028,8 +1028,8 @@ namespace Microsoft.Security.Utilities
             bool result;
             ulong seed = 42;
             string signature = "TEST";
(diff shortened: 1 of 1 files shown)


src/Tests.Microsoft.Security.Utilities.Core/IdentifiableSecretsTests.cs
```diff
@@ -1028,8 +1028,8 @@ public void IdentifiableSecrets_TryValidateBase64Key_ChecksAlphabet()
             bool result;
             ulong seed = 42;
             string signature = "TEST";
-            string secretBase64Std = "83Tv3p0dtmc/cw7eEVxcC7lh6ZM+MgPG3TESTLBLKt4=";
-            string secretBase64Url = "83Tv3p0dtmc_cw7eEVxcC7lh6ZM-MgPG3TESTLBLKt4=";
```
michaelcfanning, line 1032: >MgPG3TESTLBLKt4 [](http://example.com/codeflow?start=66&length=15)

You can formally define test keys in your specification but that doesn't mean anyone else will know about them. :)

michaelcfanning (approved):
:shipit:

PR #129, opened 2025-02-11
Fix incorrect signatures preventing rust scanner from being used for two patterns

nguerrera:
`IdentifiableScan` falls back to the slower `SecretMasker` for patterns that aren't detected by rust. Determining if rust can detect a pattern is controlled by a hard-coded list of signatures. In two cases, HIS v2 and Azure IOT, the values in the hard-coded list did not match the signatures of their `RegexPattern`s. This results in using the backup slower masker for these two patterns.

Fixing this causes tests to fail when rust fails to match `JQQJ9D` (derived) and `JQQJ9H` (hashed) patterns. This draft adds support for these to rust, but these formats never got off the ground and should probably be removed throughout instead.

Fixing this also reveals that the C# masker was including the non-alphanumeric character preceding an HIS v2 secret in its match. Once `IdentifiableScan` started to use rust for HIS v2 matching, tests failed because the C3IDs didn't match. This is fixed as well by using "refine" as the capture name instead of "secret".

Also fix missing rust release notes for previous release.

### DIFF
--- a/docs/ReleaseHistory_Rust.md
+++ b/docs/ReleaseHistory_Rust.md
@@ -13,10 +13,14 @@ This document contains release notes pertaining to the Rust crates.
 - FNS => False negative reduction in static analysis.
 
 # UNRELEASED
+- FNS: Add detection of derived and hashed HIS v2 keys for `SEC101/200: Microsoft common annotated security key (HIS v2)`.
+
+# 1.5.4 - 11/19/2024
+- DEP: Removed dependency of the `base-62` crate since it depended on the `failure` crate which has a known [vulnerability](https://github.com/advisories/GHSA-jq66-xh47-j9f3).
 - NEW: Introduce `marvin::{compute_hash_slice, compute_hash32_slice}` to compute marvin checksums directly from slices. `marvin::{compute_hash, compute_hash32}` also rely on the new, faster implementation.
 
 # 1.5.3 - 07/26/2024
-- DEP: `System.Text.Json` updated to `v8.0.4` to resolve Depandabot alert.
+- DEP: `System.Text.Json` updated to `v8.0.4` to resolve Dependabot alert.
 - NEW: Introduces the `ScanEngine` struct, which allows simplified usage in concurrent scenarios---a single `ScanEngine` instance, along with per-thread `ScanState` instances, suffice without the need for additional synchronization. The existing `Scan` struct is operationally unchanged for users.
 - BRK: `Send` and `Sync` bounds have been added for `ScanDefinition` validators. This allows `ScanEngine` to be `Send + Sync`.
 
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Security.Utilities
             Id = "SEC101/200";
             Name = nameof(CommonAnnotatedSecurityKey);
             DetectionMetadata = DetectionMetadata.Identifiable;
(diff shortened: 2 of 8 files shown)


docs/ReleaseHistory_Rust.md
```diff
@@ -13,10 +13,14 @@ This document contains release notes pertaining to the Rust crates.
 - FNS => False negative reduction in static analysis.
 
 # UNRELEASED
+- FNS: Add detection of derived and hashed HIS v2 keys for `SEC101/200: Microsoft common annotated security key (HIS v2)`.
+
+# 1.5.4 - 11/19/2024
```
nguerrera, line 18: This shipped without proper release notes. Copied below from the C# release notes.

docs/ReleaseHistory_Rust.md
```diff
@@ -13,10 +13,14 @@ This document contains release notes pertaining to the Rust crates.
...
+- DEP: Removed dependency of the `base-62` crate since it depended on the `failure` crate which has a known [vulnerability](https://github.com/advisories/GHSA-jq66-xh47-j9f3).
 - NEW: Introduce `marvin::{compute_hash_slice, compute_hash32_slice}` to compute marvin checksums directly from slices. `marvin::{compute_hash, compute_hash32}` also rely on the new, faster implementation.
 
 # 1.5.3 - 07/26/2024
-- DEP: `System.Text.Json` updated to `v8.0.4` to resolve Depandabot alert.
```
michaelcfanning, line 19: >Depandabot [](http://example.com/codeflow?start=57&length=10)

depandabot. i see.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
```diff
@@ -13,7 +13,7 @@ public CommonAnnotatedSecurityKey()
             Id = "SEC101/200";
             Name = nameof(CommonAnnotatedSecurityKey);
             DetectionMetadata = DetectionMetadata.Identifiable;
-            Pattern = $"{WellKnownRegexPatterns.PrefixBase62}(?P<secret>[{WellKnownRegexPatterns.Base62}]{{52}}JQQJ9(?:9|D|H)[{WellKnownRegexPatterns.Base62}][A-L][{WellKnownRegexPatterns.Base62}]{{16}}[A-Za-z][{WellKnownRegexPatterns.Base62}]{{7}}(?:[{WellKnownRegexPatterns.Base62}]{{2}}==)?)";
+            Pattern = $"{WellKnownRegexPatterns.PrefixBase62}(?P<refine>[{WellKnownRegexPatterns.Base62}]{{52}}JQQJ9(?:9|D|H)[{WellKnownRegexPatterns.Base62}][A-L][{WellKnownRegexPatterns.Base62}]{{16}}[A-Za-z][{WellKnownRegexPatterns.Base62}]{{7}}(?:[{WellKnownRegexPatterns.Base62}]{{2}}==)?)";
```
michaelcfanning, line 16: >H [](http://example.com/codeflow?start=123&length=1)

zap the `H`? retain the `D` for now?

src/security_utilities_rust/src/microsoft_security_utilities_core/identifiable_scans.rs
```diff
@@ -739,7 +739,7 @@ impl ScanOptions {
         let match_bytes = |data: &[u8]| -> usize {
             /*
              * Checks are equivalent to this regex:
-             * [A-Za-z0-9]{52}JQQJ99[A-Za-z0-9][A-L][A-Za-z0-9]{16}[A-Za-z][A-Za-z0-9]{7}([A-Za-z0-9]{2}==)?
+             * [A-Za-z0-9]{52}JQQJ9(9|D|H)[A-Za-z0-9][A-L][A-Za-z0-9]{16}[A-Za-z][A-Za-z0-9]{7}([A-Za-z0-9]{2}==)?
```
michaelcfanning, line 742: >H) [](http://example.com/codeflow?start=40&length=2)

zap `H`

michaelcfanning (dismissed):
:shipit:

docs/ReleaseHistory_Rust.md
nguerrera: It's common to confuse the bot that sends dependency alerts with the one that removes pandas. Naming is hard.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_CommonAnnotatedSecurityKey.cs
nguerrera: I will leave this in draft and come back to it next week after focus week and fix this and add some tests.

Issue #130, opened 2025-02-13
[PROPOSAL] : expand general classifiers for fallback tagging when precise classification fails


michaelcfanning:
In some sensitive and/or high-performance real-time scanning contexts, it may be helpful to tag or categorize data that we can't otherwise classify as actually sensitive information. This tagging may be useful in that it first might provide some confidence that a data value is not a false negative (i.e., that we can't classify it today because it can't possibly be comprise a secret). Related, this classification might help winnow/prioritize attempts to, in fact, chase down and close false negatives, where we are not detecting secrets in data streams. 

Some early analysis and thinking has generated the following list of ideas we could pursue. We should collect the broad set here as a first step and discuss an appropriate taxonomy/granularity of detection, etc., as well as think about how expanding our classifier will challenge our current design. The library today is tuned towards finding high-confidence patterns, with a corresponding assumption that multiple checks will not fire on the same data, and does not have a design that lends itself to a sieve/filtering design (in which, for example, patterns are oriented towards prioritizing high precision classifications and then falling back to less precise categorizations if appropriate.

Given the prevalence of key/value pairs in logging and other real-time detection scenarios, thinking about how/whether to incorporate an analysis of a key name is another useful direction. The ideas below refer to hints for each pattern kind that may be obtainable from this data. The items below start with a range of hints that may exist along with formats that should be packageable as a discrete regex-driven check.

- `address` : Network addresses: IPv4, IPv6
- `id`, `uuid`, `guid`: UUIDs. Historically, some security models have utilized guids as secrets but most modern providers have moved away from this approach. Detecting a UUID in a data value is a high confidence classification.
- `port`: TCP/UDP ports, constrained to 0 - 65536
- `bytes`, `seconds`, `timeout`, etc: A generic regex that detects strictly numeric values could be helpful in isolating data that appears to be configuration related or which simply is very unlikely to comprise a secret
- `account`, `user`, `id`: primary alphabetical values with some possibility of numbers to distinguish generated accounts `test1`, `test2`, etc. Many prominent account providers define mini-formats for account names that, for example, forbid that the begin with a digit, are limited in the symbols that can be included, etc. There is a possibility of creating platform/framework specific account name detections that fall back to a generic patter.
- `name`, `instance`, `resource`, `id`: Similar to previous, named instances of resources to connect to, utilize, etc.
- `machine`, `vm`, `server`: An elaboration on resource kind.
[...]

PR #131, opened 2025-02-19
feat: add initial DataClassification rules

rwoll:
Frist part of #130 implementation.

- RUL: Add `DAT101/000.GuidValue` detection as part of a new DAT101 detection series which helps classify non-sensitive data .
- RUL: Add `DAT101/001.UuidValue` non-sensitive data classification.
- RUL: Add `DAT101/002.IPv4` non-sensitive data classification.
- RUL: Add `DAT101/003.IPv6` non-sensitive data classification.
- RUL: Add `DAT101/004.Integer` non-sensitive data classification.
- RUL: Add `DAT101/005.Float` non-sensitive data classification.

Additional rules will be added in batches now that the boiler is here.


src/Microsoft.Security.Utilities.Core/ClassifiedNonSensitiveValues/SEC001_000_GuidValue.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class GuidValue : RegexPattern
+    {
+        public GuidValue()
+        {
+            Id = "SEC001/000";
```
rwoll: 1. What numeric namespace do we want all these new rules in?
2. Is the folder name appropriate? `ClassifiedNonSensitiveValues`. All other folders end with a `Keys` suffix, but all these rules are *not* `Keys`. I chose `Values` to have parallel naming consistency, but `Values` is a near-meaningless word so I'd prefer to omit entirely or open to a suggestion of a more descriptive word.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -31,13 +31,20 @@ public void SecretMasker_PreciselyClassifiedSecurityKeys_Detections()
     }
 
     // "https://github.com/microsoft/security-utilities/issues/95")
-    //[TestMethod]
+    // [TestMethod]
```
rwoll: Oops…whitspace diff. Will back this out.

src/Microsoft.Security.Utilities.Core/ClassifiedNonSensitiveValues/SEC001_000_GuidValue.cs
rwoll: `NonSensitive` may also be an overreach. Just because it's a GUID, doesn't mean someone is not using it as a password/API key.

src/Microsoft.Security.Utilities.Core/ClassifiedNonSensitiveValues/SEC001_000_GuidValue.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            // TODO(rosswollman): The test harness does not have a formal way to name test cases and debug specific test cases. (e.g. what Theory or other data driven tests would provide out of the box.)
+            // Test Case Name: MixedCaseAndAlphaNumerics
+            yield return "12Aa0000-0Cd0-00eF-cB00-eF0000000000";
+            yield return Guid.NewGuid().ToString();
+        }
+
+        // TODO(rosswollman): The repo does not have tests for True Negatives.
```
rwoll: GenerateFalsePositiveExamples.

src/Microsoft.Security.Utilities.Core/ClassifiedNonSensitiveValues/SEC001_000_GuidValue.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Security.Utilities
+{
+    public class GuidValue : RegexPattern
+    {
+        public GuidValue()
+        {
+            Id = "SEC001/000";
+            Name = nameof(GuidValue);
+            Pattern = @"(?i)^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$";
```
rwoll: From Mike, use this more specific regex instead:

```
(?i)^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$
```

src/Microsoft.Security.Utilities.Core/ClassifiedNonSensitiveValues/SEC001_000_GuidValue.cs
rwoll: Rule namespace: `DAT101/*`
Directory Name: `DataClassification`

rwoll (commented):
Adding some comments from @michaelcfanning during sync review. I'll incorporate into next pass.

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,12 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- NEW: Introduce new `DAT101` detection series which help users classify likely-non-secret values into more specific data types (e.g. IP addresses, GUIDs, timeouts, etc.).
```
rwoll: FYI, `NEW` is intentionally out of order from the prescribed release note ordering list since it makes more sense to introduce the new "feature" set *before* enumerating all the rules.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_FormalGuid.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            // This pattern encodes a more formal definition of a GUID compared to what might colloquially be thought of as a GUID.
+            //
+            // Specifically, this pattern ensures there's a 1-5 version identifier in the 3rd block and that the 4th block is constrained to the GUID variant.
+            // .NET's System.Guid.Parse will accept more than just this format, but Guid.NewGuid() will always generate a v4 GUID.
+            //
+            // There are many organic examples of GUIDs that do not meet this strict definition:
+            // * 3.5M results on GitHub: https://github.com/search?q=%2F%5B0-9a-f%5D%7B8%7D-%5B0-9a-f%5D%7B4%7D-%5B0-9a-f%5D%7B4%7D-%5B0-9a-f%5D%7B4%7D-%5B0-9a-f%5D%7B12%7D%2F%20NOT%20%2F%5B0-9a-f%5D%7B8%7D-%5B0-9a-f%5D%7B4%7D-%5B1-5%5D%5B0-9a-f%5D%7B3%7D-%5B89ab%5D%5B0-9a-f%5D%7B3%7D-%5B0-9a-f%5D%7B12%7D%2F%20&type=code
+            // * Many of the Microsoft-verified first party app ids (e.g.Office 365 SharePoint Online): https://learn.microsoft.com/en-us/troubleshoot/entra/entra-id/governance/verify-first-party-apps-sign-in
+            Pattern = @"(?i)^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$";
```
rwoll: @michaelcfanning - Can you help add a comment here articulating *why* we are choosing a more restrictive GUID pattern than the originally proposed?

Admittedly, I thought I understood your reasoning yesterday, but I'm struggling to articulate it so I may not have understood the reasoning as much as I thought. :)

Since there's intentionality behind the choice, I want to ensure we've documented it for our future selves inline.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_001_Ip4.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class Ip4 : RegexPattern
```
michaelcfanning: >Ip4 [](http://example.com/codeflow?start=17&length=3)

NIT, Ip4 does not meet .net casing standard, where IP4 would.

Maybe we could use `IPv4`? This is the term I've mostly seen, it's the title for this. e.g., [https://en.wikipedia.org/wiki/IPv4](https://en.wikipedia.org/wiki/IPv4)

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_001_Ip4.cs
```diff
@@ -0,0 +1,28 @@
...
+        public Ip4()
+        {
+            Id = "DAT101/001";
+            Name = nameof(Ip4);
+            Pattern = @"^((25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\.){3}(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])$";
+        }
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            yield return "127.0.0.1";
```
michaelcfanning: >127.0.0.1 [](http://example.com/codeflow?start=26&length=9)

I suppose there's no utility in pulling in IPAddress for any of this, yes?

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_002_Ip6.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            Id = "DAT101/002";
+            Name = nameof(Ip6);
+            Pattern = @"(?i)^([0-9a-f]{1,4}:){7}[0-9a-f]{1,4}$";
+        }
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            var example = "3fff:fff:ffff:ffff:ffff:ffff:ffff:ffff";
+            yield return example.ToLowerInvariant();
```
michaelcfanning: >ToLowerInvariant [](http://example.com/codeflow?start=33&length=16)

example is already lower-case, why generate a short-lived string that duplicates the test case?

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_002_Ip6.cs
```diff
@@ -0,0 +1,29 @@
...
+            yield return example.ToUpperInvariant();
+        }
+
+        public override IEnumerable<string> GenerateFalsePositiveExamples()
+        {
+            // zero-compression is not handled by this rule
```
michaelcfanning: >zero [](http://example.com/codeflow?start=15&length=4)

Our comments s/be well-formed sentences so upper case `Zero-compression` and add a period at the end of the sentence.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_002_Ip6.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class Ip6 : RegexPattern
```
michaelcfanning: >Ip6 [](http://example.com/codeflow?start=17&length=3)

`IPv6` instead?

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_003_Integer.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Security.Utilities
+{
+    public class Integer : RegexPattern
+    {
+        public Integer()
+        {
+            Id = "DAT101/003";
+            Name = nameof(Integer);
+            Pattern = @"^-?\d+$";
```
michaelcfanning, line 13: >-? [](http://example.com/codeflow?start=25&length=2)

ha. nice catch. 

I suppose a prefixed `+` is not required. If we did add it we would create base64 confusion! #Resolved

michaelcfanning: but what about decimal values? hm. what about locale-rendered formatted numbers?

12.0
1,234,500

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_004_Float.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class Float : RegexPattern
+    {
+        public Float()
```
michaelcfanning, line 9: >Float [](http://example.com/codeflow?start=15&length=5)

ah! you added float.

your regex looks good, though I'd add a case with an explicit leading zero for the negative case, i.e., `-0.5`

also for this rule, should we attempt any parsing?

note that this rule won't find `12,5` in locales where the comma is the locale-specific decimal separator. someone somewhere must have made a mistake by configuring data in a culture-sensitive locale!

just thinking aloud, we can always ship some basics and come back to the topic.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_FormalGuid.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class FormalGuid : RegexPattern
```
michaelcfanning: >FormalGuid [](http://example.com/codeflow?start=17&length=10)

A formal GUID on research turns out to be a `UUID`

so let's rename this `Uuid` and we can create another rule for that apparent `Guid` spillover.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_FormalGuid.cs
```diff
@@ -0,0 +1,40 @@
...
+            // There are many organic examples of GUIDs that do not meet this strict definition:
+            // * 3.5M results on GitHub: https://github.com/search?q=%2F%5B0-9a-f%5D%7B8%7D-%5B0-9a-f%5D%7B4%7D-%5B0-9a-f%5D%7B4%7D-%5B0-9a-f%5D%7B4%7D-%5B0-9a-f%5D%7B12%7D%2F%20NOT%20%2F%5B0-9a-f%5D%7B8%7D-%5B0-9a-f%5D%7B4%7D-%5B1-5%5D%5B0-9a-f%5D%7B3%7D-%5B89ab%5D%5B0-9a-f%5D%7B3%7D-%5B0-9a-f%5D%7B12%7D%2F%20&type=code
+            // * Many of the Microsoft-verified first party app ids (e.g.Office 365 SharePoint Online): https://learn.microsoft.com/en-us/troubleshoot/entra/entra-id/governance/verify-first-party-apps-sign-in
+            Pattern = @"(?i)^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$";
+        }
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            var azureCliClientGuid = "04b07795-8ddb-461a-bbee-02f9e1bf7b46";
+            yield return azureCliClientGuid.ToLowerInvariant();
```
michaelcfanning: >ToLowerInvariant [](http://example.com/codeflow?start=44&length=16)

your test case is already lower invariant.

your test cases don't include all version values, we should explicitly declare examples for v 1,2,3,4 (which you have),5, 7, and 8

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_FormalGuid.cs
```diff
@@ -0,0 +1,40 @@
...
+            yield return azureCliClientGuid.ToUpperInvariant();
+            yield return Guid.NewGuid().ToString();
+        }
+
+        public override IEnumerable<string> GenerateFalsePositiveExamples()
+        {
+            // Office 365 SharePoint Online App Id - https://learn.microsoft.com/en-us/troubleshoot/entra/entra-id/governance/verify-first-party-apps-sign-in
```
michaelcfanning: >Office 365 SharePoint Online App Id [](http://example.com/codeflow?start=15&length=35)

More precisely, this value is actually a COM programming GUID, not a UUID.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_001_Ip4.cs
rwoll: Agreed, let's go with IPv4 and IPv6, matching casing (and also what I typically use) from .NET: https://learn.microsoft.com/en-us/dotnet/api/system.net.ipaddress.ipv6any?view=net-9.0#system-net-ipaddress-ipv6any

Darn `Url` and `IP` casing being inconsistent in .NET!

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_001_Ip4.cs
rwoll: Not yet. For broad buckets, I think we're good without the additional processing for now.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_002_Ip6.cs
rwoll: Intent was to make it more clear that one case intnetionall tests lower while the other is upper.

The bespoke testing model this repo uses does not have a way to provide test case names.

Since your comment indicates the code's intent was not communicated clearly, I'll just remove the upper and lower calls and inline some values with a title-as-comment.

docs/ReleaseHistory.md
michaelcfanning: If you do this, then you need to change the legend to conform to your preferred ordering and you'd need to update all the release notes as well.

Personally, I think your `NEW` note is simply redundant, just add the new checks. I'd handle this by making the first entry a bit more built-out then add the remainder.

just spitballing:

- RUL: Add `DAT101/000.Uuid` detection as part of a new `DAT101` detection series which helps classify non-sensitive data .
- RUL: Add `DAT101/001.Ip4` non-sensitive data classification.
- RUL: Add `DAT101/002.Ip6`  non-sensitive data classification.
- RUL: Add `DAT101/003.Integer`  non-sensitive data classification.
- RUL: Add `DAT101/004.Float`  non-sensitive data classification.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -92,6 +92,15 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
         new AzureEventGridIdentifiableKey(),
     };
 
+    public static IEnumerable<RegexPattern> DataClassification { get; } = new RegexPattern[]
+    {
+        new FormalGuid(),
```
michaelcfanning: >FormalGuid [](http://example.com/codeflow?start=12&length=10)

this order is neither alphabetical (one possible consistent ordering standard) nor waterfalled by length (a second ordering standard).

If you take all my notes, here's how I would have organized these checks:

        new IPv4(),
        new IPv6(),
        new Guid(),
        new Uuid(),
        new Float(),
        new Integer(),

docs/ReleaseHistory.md
rwoll: Sg. I'll remove it and use your suggestion to inline it in the first bullet.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
rwoll: Ah, I ordered based on Rule #.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_FormalGuid.cs
rwoll: noted. renamed, and will breakout another class. Thanks!

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_004_Float.cs
rwoll: I'll add a specific negative test case and comment (and issue), and we can revisit. It's a very excellent point though, and I'd be interested to run the classifier to see how common seeing these localized variants are in the data of interest.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_003_Integer.cs
rwoll: Decimal is covered by float rule. Localization we will punt on for now: https://github.com/microsoft/security-utilities/pull/131#discussion_r1964381629

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_002_Ip6.cs
rwoll: Updated to make more clear/explicit and do away with the trickery!

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_FormalGuid.cs
rwoll: Comment is moot. We will have two rules: GUID vs. UUID.

src/Tests.Microsoft.Security.Utilities.Core/RegexPatternTests.cs
```diff
@@ -48,7 +48,7 @@ public void RegexPattern_Equals_ReturnsTrue_WhenIdsDiffer()
     {
         // Arrange
         var secret1 = new RegexPattern(Id, Name, DetectionMetadata.Identifiable, "abc");
-        var secret2 = new RegexPattern($"{Guid.NewGuid()}", Name, DetectionMetadata.Identifiable, "abc");
+        var secret2 = new RegexPattern($"{System.Guid.NewGuid()}", Name, DetectionMetadata.Identifiable, "abc");
```
rwoll: Let's change our RegexPattern `Guid` to `GuidValue` to avoid the name clashes.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_001_UuidValue.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            // Example of a UUIDv6 Value from https://www.rfc-editor.org/rfc/rfc9562.html#name-example-of-a-uuidv6-value.
+            yield return "1EC9414C-232A-6B00-B3C8-9F6BDECED846";
+
+            // Example of a UUIDv7 Value from https://www.rfc-editor.org/rfc/rfc9562.html#name-example-of-a-uuidv7-value.
+            yield return "017F22E2-79B0-7CC3-98C4-DC0C0C07398F";
+
+            // Example of a UUIDv8 Value (Time-Based) from https://www.rfc-editor.org/rfc/rfc9562.html#name-example-of-a-uuidv8-value-t.
+            yield return "2489E9AD-2EE2-8E00-8EC9-32D5F69181C0";
+        }
```
rwoll, line 51: NB: our tests will ensure the Guid TruePositives won't fire this rule.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.Security.Utilities
+{
+    public class GuidValue : RegexPattern
+    {
+        public GuidValue()
+        {
+            Id = "DAT101/000";
+            Name = nameof(GuidValue);
+            Pattern = @"(?i)^([0-9a-f]{8}-[0-9a-f]{4}-[0|9|a|b|c|d|e|f][0-9a-f]{3}-[0-9a-f]{4}-[0-9a-f]{12})" +
```
rwoll, line 14: Code has been updated to serpate out UUIDs from GUIDs.

As a consequence, everything about the code has grown in complexity and cognitive load.

We should articulate *why* its important to seperate the two rules in the form of an inline comment. Otherwise, we should strongly consider smooshing back into one straightforward Guid rule. @michaelcfanning - can you provide that inline comment? I'm struggling to rationalize our choice here. On some principled level, I understand, but I'm not sure the tradeoffs warrant the division of rules. Hopefully a concise, thoughtful inline comment can help there.

One major concern I have is we will now miss Guids by trying to have mutually exclusive rules, and if this mutual exclusivity is implemented incorrectly, we will miss swaths of classifications.

rwoll (commented):
okie - ready for another review pass.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
nguerrera: I am late to this whole discussion, but I must admit I don't understand the value of mutually exclusive rules. Who is the customer that needs this?

I thought the purpose of these was to categorize things that are not likely secrets. Do we really need this level of detail on exactly what kind of UUID/GUID we have for that categorization? This sort of micro-categorization seems outside the domain of a tool based on regex matching to me.  

Also, there are perf costs to running two rules instead of 1 when all you care about is the union of the two rules, which seems like the far more likely use case here.

rwoll: I'll defer to @michaelcfanning on articulating the reason for seperating. I loosely follow the rationale to seperate. Admittedly, not enough to properly articulate it.

src/Tests.Microsoft.Security.Utilities.Core/RegexPatternTests.cs
rwoll: Even though this is less of a concern with Uuid, I chose parallel name (`UuidValue` vs. `Uuid`)

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            return base.GetMatchIdAndName(match);
+        }
+
+        public override IEnumerable<string> GenerateTruePositiveExamples()
+        {
+            // Example of a GUID, but not a UUID. This example is the Office 365 SharePoint Online App Id. See https://learn.microsoft.com/en-us/troubleshoot/entra/entra-id/governance/verify-first-party-apps-sign-in.
+            yield return "00000003-0000-0ff1-ce00-000000000000";
+
+            // Example of almost a UUIDv4, but the first 'c' of the 4th block ('cbac') makes it not a UUID, but a GUID.
```
michaelcfanning, line 33: >UUIDv4 [](http://example.com/codeflow?start=35&length=6)

this comment is wrong, it is the second 4-char sequences that encodes the version, which here is 4 so why are we matching?

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
```diff
@@ -0,0 +1,46 @@
...
+            yield return "919108f7-52d1-4320-cbac-f847db4148a8";
+        }
+
+        public override IEnumerable<string> GenerateFalsePositiveExamples()
+        {
+            // Example of a UUIDv4 Value from https://www.rfc-editor.org/rfc/rfc9562.html#name-example-of-a-uuidv4-value.
```
michaelcfanning, line 39: >UUIDv4 [](http://example.com/codeflow?start=28&length=6)

note your comment here is correct, the 2nd 4 char sequence has v4

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
michaelcfanning: Oh geez. Well, I have backed us into this complexity, yes. For one thing, there is no over-arching term that covers both GUIDs and UUIDs. `SpeciallyFormatted128BitValue` ?

So my thought is that Guid is useful to concretely separate data that isn't conclusively specified as meeting the randomness guarantees specified in UUID production. Like the IUnknown GUID. These constructs, which can simply be well-documented, preallocated, non-random 128 bit values should never be used/presented as secrets. This is distinct from UUIDs where this generated has been used as secrets, e.g., in older AAD SPN password generation and some historical Azure OpenAI keys.

From this angle, what we're calling a GUID has an arguably higher level of certainty over other data to be non-sensitive. Going through all this, though, not sure the distinction is helping us.

If we want to simplify, let's drop the `000` rule (we shouldn't index starting at zero in any case). And just call it a GuidValue to reflect our MS orientation (where the .NET class is permissive accepting all 128 bits values).

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
michaelcfanning: You didn't create this technical debt but generally we should be encoding the readable name every time we refer to a friendly name. In this case, devs can just f12 into the thing though....

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_004_Float.cs
michaelcfanning: you didn't add my test case of `-0.5` but I am not super worried about it.

michaelcfanning (approved):
:shipit:

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
michaelcfanning, line 33: >cbac [](http://example.com/codeflow?start=80&length=4)

ah, so this little wrinkle really point to the nuance you mentioned. but honestly, if GUIDs and UUIDs are all merged together, how the heck are you supposed to distinguish 'proper' UUIDs from, say, a predefined GUID where someone has chosen to share a common value for this reserved bit?

rwoll:
Merged but will responde to the comments and follow up with PRs depending on outcomes.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
rwoll: >  For one thing, there is no over-arching term that covers both GUIDs and UUIDs. SpeciallyFormatted128BitValue ?

I think all UUIDs are GUIDs. Not all GUIDs are UUIDs. Regardless, colloquially GUID is what many people use to  refer to "SpeciallyFormatted128BitValue" as.

> So my thought is that Guid is useful to concretely separate data that isn't conclusively specified as meeting the randomness guarantees specified in UUID production.

If I understand correctly, this claims that b/c something looks like a UUID based solely on 2 characters, we should conclude it has likely been generated with some amount of randomness. I don't think this is a valid claim so any distinction on the grounds of these 2 characters is actually misleading.

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_000_GuidValue.cs
rwoll: > how the heck are you supposed to distinguish 'proper' UUIDs from, say, a predefined GUID where someone has chosen to share a common value for this reserved bit?

You don't distinguish them. Their usage and generation cannot be confidently inferred nor claimed from observing the ouput in the wild.

PR #134, opened 2025-02-22
chore: remove Uuid rule

rwoll:
Follow up to #131.

Removes dedicated Uuid classification. For the purposes of this tool, seperating what appear to be Uuids vs. more generic Guids, is unhelpful to the core goal. Hence removing to vastly simplify implementation, avoid misleading labels, and decrease cognitive load for folks consuming these labels.

See [related discussion](https://github.com/microsoft/security-utilities/pull/131#discussion_r1964496500).


michaelcfanning (dismissed):
:shipit:

src/Microsoft.Security.Utilities.Core/DataClassification/DAT101_001_GuidValue.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public class GuidValue : RegexPattern
+    {
+        public GuidValue()
+        {
+            Id = "DAT101/000";
```
rwoll:
```suggestion
            Id = "DAT101/001";
```

michaelcfanning (approved):
:shipit:

PR #139, opened 2025-02-25
Drop net451 and net6.0 support

nguerrera:
This change is split off from #132 to allow other work that relies on reduced framework support to happen in parallel to closing on #132. Also, it really should have been its own PR anyway. :)

### DIFF
--- a/src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/SecurityUtilitiesApiUtilizationExample.csproj
+++ b/src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/SecurityUtilitiesApiUtilizationExample.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-	  <TargetFrameworks>net451;net462;net6.0</TargetFrameworks>
+	  <TargetFrameworks>net8.0;net461</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


rwoll (approved):
🚀 it!

src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/SecurityUtilitiesApiUtilizationExample.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-	  <TargetFrameworks>net451;net462;net6.0</TargetFrameworks>
+	  <TargetFrameworks>net8.0;net461</TargetFrameworks>
```
michaelcfanning, line 5: >net461 [](http://example.com/codeflow?start=28&length=6)

i think we have a consumer of this package that targets `net472` FYI, i.e., this is the minimal support i'm aware of that we need

nguerrera: I actually did spot a net461 in the wild that references the package. I'll follow up on that offline. It so happens that net461 is the minimum version that supports netstandard2.0 (with some caveats), that's why I chose that for this test.

PR #148, opened 2025-02-27
Validate and standardize named captures and optimize cached regex lookup

nguerrera:
- Patterns must have at most one named capture and it must be named 'refine'. Enforce this in a unit test and fix the resulting issues.
- Standardize on `?<refine>` over `?P<refine>` syntax. This is the standard of the vast majority of patterns and is supported by both RE2 and .NET. Engines that require 'P' such as Python would already have had to add it to most of the patterns.
- Remove capture group normalization on GetOrCreateRegex. This is unnecessary after this standardization.
- Use a value tuple to eliminate allocation on regex lookup.
- Prevent closure allocation by using the argument passed to GetOrAdd callback instead of retrieving it from closure.
- Remove explicit static constructor so that CachedDotNetRegex is emitted with `beforefieldinit` flag, which permits better native code gen that no longer has to ensure the precise timing of static initialization.
- Reduce iterations of IdentifiableScan_IdentifiableKeys test to 1. The value of 1000 was meant for local debugging only.

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -11,6 +11,12 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- BRK: Regular expression syntax has been standardized in JSON to conform to how the overwhelming majority of patterns were already defined.
+  - `refine` is used now used throughout as the name of the capture group used to isolate an actual find from the full expression that also matches delimiting characters. `secret` was previously used in some instances.
+  - `?<name>` is now used throughout for named captures. '?P<name>' was previously used in some instances. This may require replacing '?<' with '?P<' if using a regex engine that only accepts the '?P<name>' syntax.
+- BRK: `CachedDotNetRegexEngine`will no longer accept `(?P<name>)` syntax. This is only relevant if it is used with patterns other than those distributed with this library.
+- BUG: `SEC101/200.CommonAnnotatedSecurityKey` considered non-alphanumeric delimiter preceding secret to be part of the match.
 # 1.14.0 - 02/25/2025
 - RUL: Add `DAT101/001.GuidValue` detection as part of a new DAT101 detection series which helps classify non-sensitive data.
 - RUL: Add `DAT101/002.IPv4` non-sensitive data classification.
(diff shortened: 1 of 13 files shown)


docs/ReleaseHistory.md
```diff
@@ -11,6 +11,12 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- BRK: Regular expression syntax has been standardized in JSON to conform to how the overwhelming majority of patterns were already defined.
+  - `refine` is used now used throughout as the name of the capture group used to isolate an actual find from the full expression that also matches delimiting characters. `secret` was previously used in some instances.
```
michaelcfanning, line 16: >refine [](http://example.com/codeflow?start=5&length=6)

I hate to mention this but citing the specific rules that were subject to the changes you've made would be good documentation to have here. #Resolved

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,12 @@
...
+  - `?<name>` is now used throughout for named captures. '?P<name>' was previously used in some instances. This may require replacing '?<' with '?P<' if using a regex engine that only accepts the '?P<name>' syntax.
+- BRK: `CachedDotNetRegexEngine`will no longer accept `(?P<name>)` syntax. This is only relevant if it is used with patterns other than those distributed with this library.
+- BUG: `SEC101/200.CommonAnnotatedSecurityKey` considered non-alphanumeric delimiter preceding secret to be part of the match.
```
michaelcfanning: >CommonAnnotatedSecurityKey [](http://example.com/codeflow?start=19&length=26)

FWIW, this rule name is `UnclassifiedCommonAnnotatedSecurityKey` in other current project work. Not your bug, of course, a net new suggested change. #Resolved

michaelcfanning: >match [](http://example.com/codeflow?start=120&length=5)

NIT introduce line break to separate unreleased from 1.14 #Resolved

michaelcfanning (dismissed):
:shipit:

docs/ReleaseHistory.md
nguerrera: So I think this means I should leave it and let the change that renames it, rename it here too?

docs/ReleaseHistory.md
nguerrera: Done.

nguerrera:
@michaelcfanning Can you take another look? I addressed the feedback and this needs a re-stamp.

docs/ReleaseHistory.md
nguerrera: Done.

michaelcfanning (approved):
:shipit:

PR #149, opened 2025-02-27
Remove incorrect signatures from SEC101/061.LooseOAuth2BearerToken

nguerrera:
Also re-enable the test that would have caught this but was disabled against #95. Instead of disabling the test wholesale, add a way to tolerate multiple findings for these rules, still linking to #95 as the reason.

### DIFF
--- a/GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json
+++ b/GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json
@@ -34,10 +34,7 @@
     "Pattern": "(?i)authorization:(\\s|%20)bearer(\\s|%20)(?<refine>[0-9a-z][abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_~.\\-+\\/=]*)([^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_~.\\-+/=]|$)",
     "Id": "SEC101/061",
     "Name": "OAuth2BearerToken",
-    "Signatures": [
-      "sig=",
-      "ret="
-    ],
+    "Signatures": null,
     "DetectionMetadata": "LowConfidence"
   },
   {
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -23,7 +23,7 @@
       - `SEC101/565.SecretScanningSampleToken`
 - BRK: `CachedDotNetRegexEngine`will no longer accept `(?P<name>)` syntax. This is only relevant if it is used with patterns other than those distributed with this library.
 - BUG: `SEC101/200.CommonAnnotatedSecurityKey` and `SEC101/565.SecretScanningSampleToken` considered non-alphanumeric delimiter preceding secret to be part of the match.
-- BUG  `SEC101/061.LooseOAuth2BearerToken` had incorrect signatures, causing no matches to be found unless the input happened to also contain `sig=` or `ret=`
+- BUG  `SEC101/061.LooseOAuth2BearerToken` had incorrect signatures, causing no matches to be found unless the input happened to also contain `sig=` or `ret=`.
 
 # 1.14.0 - 02/25/2025
 - RUL: Add `DAT101/001.GuidValue` detection as part of a new DAT101 detection series which helps classify non-sensitive data.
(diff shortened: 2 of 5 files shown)


docs/ReleaseHistory.md
```diff
@@ -23,6 +23,7 @@
       - `SEC101/565.SecretScanningSampleToken`
 - BRK: `CachedDotNetRegexEngine`will no longer accept `(?P<name>)` syntax. This is only relevant if it is used with patterns other than those distributed with this library.
 - BUG: `SEC101/200.CommonAnnotatedSecurityKey` and `SEC101/565.SecretScanningSampleToken` considered non-alphanumeric delimiter preceding secret to be part of the match.
+- BUG  `SEC101/061.LooseOAuth2BearerToken` had incorrect signatures, causing no matches to be found unless the input happened to also contain `sig=` or `ret=`
```
```suggestion
- BUG  `SEC101/061.LooseOAuth2BearerToken` had incorrect signatures, causing no matches to be found unless the input happened to also contain `sig=` or `ret=`.
```

GeneratedRegexPatterns/UnclassifiedPotentialSecurityKeys.json
```diff
@@ -34,10 +34,7 @@
     "Pattern": "(?i)authorization:(\\s|%20)bearer(\\s|%20)(?<refine>[0-9a-z][abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_~.\\-+\\/=]*)([^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890_~.\\-+/=]|$)",
     "Id": "SEC101/061",
     "Name": "OAuth2BearerToken",
-    "Signatures": [
-      "sig=",
-      "ret="
-    ],
+    "Signatures": null,
```
michaelcfanning, line 37: >null [](http://example.com/codeflow?start=18&length=4)

it doesn't make sense to have no signature here if the regex itself is encoding things like `authorization` and `bearer`. either one would be ok, the choice would depend on what we think would be less likely to occur in other data. this format, oauth2 is a real bear due to the many variations in the way the secret is expressed and variance in the literal formats of the embedded secrets. blah.

nguerrera: Hmm, can I make the signature case-insensitive?

michaelcfanning: [https://www.rfc-editor.org/rfc/rfc6750#section-2.1](https://www.rfc-editor.org/rfc/rfc6750#section-2.1)

michaelcfanning: [https://auth0.com/blog/the-bearer-token-case/](https://auth0.com/blog/the-bearer-token-case/)

michaelcfanning: OK, so the RFC is ABNF and that string literal is case-insensitive, it appears, but there are suggestions that in the wild, there is a tendency to require the `Bearer` casing.

We could add `Bearer` and `bearer` as signatures and probably be ok.

rwoll: While various specs claim to be case-sensitive, in practice most server header handlers and auth middleware treat the `Authorization` header name and then the `Bearer` keyword within its value case insensitively. Here's an example of ASP.NET JwtBearer logic: https://github.com/dotnet/aspnetcore/blob/8b77ff5251b9a90b9a937330bd27bfaf71b701f4/src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs#L84

curl'ing a few different popular services confirm :)

michaelcfanning: very good, but the question is whether case insensitivity is required in this sniff.

i am changing my suggestion, which may still be rejected, use `earer` as a sniff to cover the all lowercase and Pascal-Cased forms.

this rule is not in super great shape in any case, by the way, we could use a pass over it. all these other rules don't have a sniff either, so we could simply use null as they do. if this rule were to be improved to a point where the confidence was higher, the need for a sniff would become more important.

nguerrera: > all these other rules don't have a sniff either, so we could simply use null as they do

This seems the most sensible to me.

> we could use a pass over it. 

I could file a follow-up on this.

nguerrera: #150

nguerrera: Can I get a stamp?

michaelcfanning (approved):

PR #157, opened 2025-02-28
Ado pat format exception

michaelcfanning:
- BUG: Resolve `System.FormatException: The input is not a valid Base-46 string` errors calling `IdentifiableSecrets.ValidateChecksum` with invalid base64. The API now returns `false` in this case.
- BUG: Resolve 'System.NullReferenceException` on calling `RegexPattern.GetMatchMoniker` when its internal call to `GetMatchIdAndName` return null. A null return from `GetMatchIdAndName` is an expected value that indicates post-processing has determined there is no actual match.
- BUG: Resolve `System.FormatException: The input is not a valid Base-46 string` errors calling `SEC101/102.AdoPat.GetMatchIdAndName` with invalid base64. The API now returns null in this case (the standard behavior when post-processing does not detect a match).
- BRK: `IdentifiableSecrets.ComputeDerivedCommonAnnotatedKey` now exclusive throws `System.ArgumentException` for invalid key inputs (no longer raising `System.FormatException: The input is not a valid Base-46 string` for invalid data).

See bug list above. This change also opportunistically removes a couple of places where tests ran 1000 iterations (a debug only operation). I will go rip out all these knobs in a single PR next.

This whole, little rat's nest of bugs were revealed on exercising some partner functional tests. These partner tests generated synthetic test patterns and attempted to resolve the pattern ids, if possible, directly from test examples, specifically by invoking a moniker creation api that is a simple wrapper on `GetMatchIdAndName`.

Two important issues with this:
- By attempting to post-process raw test patterns, some data passed to the helpers itself failed validation due to not having benefit of the refinement provided during detection.
- More significantly, `GetMatchMoniker` is a simple helper on top of `GetMatchIdAndName` and API that by design returns null to indicate it has been passed data deemed to be a false positive. `GetMatchMoniker` is now hardened to receive and pass along the expected null result for false positives.

One of the invalid tests patterns mentioned in the first bullet point revealed that the ADO Pat checksum validation is subject to raising `FormatException` when passed invalid base64, so this change hardens that rule-specific code path.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -765,6 +765,11 @@ public static class IdentifiableSecrets
     {
         bytes = ConvertFromBase64String(key);
 
+        if (bytes == null)
+        {
+            return false;
+        }
+
         byte[] checksumBytes = new byte[4];
         Array.Copy(bytes, bytes.Length - checksumSizeInBytes, checksumBytes, 0, checksumSizeInBytes);
 
@@ -995,7 +1000,18 @@ public static class IdentifiableSecrets
     {
         text = TransformToStandardEncoding(text);
         text += RetrievePaddingForBase64EncodedText(text);
-        return Convert.FromBase64String(text);
+        byte[] result;
+
+        try
+        {
+            result = Convert.FromBase64String(text);
+        }
+        catch
+        {
+            return null;
+        }
+
+        return result;
     }
 
     private static string ConvertToBase64String(byte[] data, bool encodeForUrl)
--- a/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
+++ b/src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
@@ -763,10 +763,13 @@ public static class IdentifiableSecrets
 
     public static bool ValidateChecksum(string key, ulong checksumSeed, out byte[] bytes)
     {
-        bytes = ConvertFromBase64String(key);
-
-        if (bytes == null)
+        try
         {
+            bytes = ConvertFromBase64String(key);
+        }
+        catch (FormatException)
+        {
+            bytes = null;
             return false;
         }
 
@@ -1002,14 +1005,7 @@ public static class IdentifiableSecrets
         text += RetrievePaddingForBase64EncodedText(text);
         byte[] result;
 
-        try
-        {
-            result = Convert.FromBase64String(text);
-        }
-        catch
-        {
-            return null;
-        }
+        result = Convert.FromBase64String(text);
 
         return result;
     }
--- a/src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
+++ b/src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Security.Utilities
         };
 
         [TestMethod]
-        public void WellKnownRegexPatterns_RunAllTheThings()
+        public void WellKnownRegexPatterns_GetMatchMonikerHardenedForOutOfOrderExecution()
         {
             using var assertionScope = new AssertionScope();
 
@@ -41,14 +41,32 @@ namespace Microsoft.Security.Utilities
                                           generateCorrelatingIds: true,
                                           RE2RegexEngine.Instance);
 
-            bool result;
-
             foreach (var pattern in patterns)
             {
                 foreach (string example in pattern.GenerateTruePositiveExamples())
                 {
-                    result = masker.DetectSecrets(example).Any() != default;
+                    // It is not obvious, but this is a test to ensure this call
+                    // does not raise exceptions. This API calls into 
+                    // 'GetMatchIdAndName' which itself may return null, indicating
+                    // that post-processing has determined that the pattern is not
+                    // a match. Because we are in the context of producing positive
+                    // test cases, this means that the test pattern contains data
+                    // that itself will be removed on the preliminary match operation.
+                    // We will therefore call this code agains to ensure that it
+                    // is no longer null post-detection.
+                    string moniker = pattern.GetMatchMoniker(example);
+
+                    var detection = masker.DetectSecrets(example).FirstOrDefault((d) => d.Id == pattern.Id);
+                    
+                    bool result = detection != default;
                     result.Should().BeTrue(because: $"Pattern '{pattern.GetType().Name}' should match '{example}'");
+
+                    if (moniker == null)
+                    {
+                        string matched = example.Substring(detection.Start, detection.End - detection.Start);
+                        moniker = pattern.GetMatchMoniker(matched);
+                        moniker.Should().NotBeNull(because: $"'{matched}' should produce a non-null moniker for {pattern.GetType().Name}' test data");
+                    }
                 }
             }
         }
--- a/src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
+++ b/src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
@@ -58,8 +58,18 @@ namespace Microsoft.Security.Utilities
 
                     var detection = masker.DetectSecrets(example).FirstOrDefault((d) => d.Id == pattern.Id);
                     
-                    bool result = detection != default;
-                    result.Should().BeTrue(because: $"Pattern '{pattern.GetType().Name}' should match '{example}'");
+                    // Currently, some rules are tuned not to double-fire, i.e., they determine
+                    // whether a separate rule might identify a pattern and, if so, the rule 
+                    // drops the result. This is a problematic design. For one thing, we don't
+                    // to see if the more precise rule is enabled. This gives the appearance of
+                    // false negatives in these low-level checks. This is a subtle topic
+                    // potentially arguing for redesign of the engine or our test expectations.
+                    bool result = 
+                        detection != default ||
+                        pattern.Name == nameof(Unclassified32CharacterString) ||
+                        pattern.Name == nameof(Unclassified16ByteHexadecimalString);
+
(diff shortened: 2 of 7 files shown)


src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -74,11 +129,10 @@ public void WellKnownRegexPatterns_EnsureAllPatternsExpressConfidence()
             }
         }
 
-
         [TestMethod]
         public void WellKnownRegexPatterns_EnsureAllMediumConfidenceOrBetterPatternsDefineSignatures()
         {
-            for (int i = 0; i < 1000; i++)
+            for (int i = 0; i < 1; i++)
```
rwoll, line 135: In your next iteration, consider removing the for loop entirely. The IDE's "Run Until Failure" can serve this usecase well. I do appreciate not doing too much refctoring here to keep the diff clean. #Pending

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,10 +26,65 @@ public class WellKnownRegexPatternsTests
             "SEC101/109.AzureContainerRegistryLegacyKey"
         };
 
+        [TestMethod]
```
rwoll, line 29: what's the code coverage like for all the changed code? This looks only match moniker related. Do the changes in `IdentifiableSecretsTests` cover all the new exception code paths? My gut says to expect more tests given the number of places wrapped in try/catch.

Though maybe your comment within this test implies the coverage is all there? #Resolved

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,10 +26,65 @@ public class WellKnownRegexPatternsTests
...
+                    string moniker = pattern.GetMatchMoniker(example);
+
+                    var detection = masker.DetectSecrets(example).FirstOrDefault((d) => d.Id == pattern.Id);
+                    
+                    // Currently, some rules are tuned not to double-fire, i.e., they determine
+                    // whether a separate rule might identify a pattern and, if so, the rule 
+                    // drops the result. This is a problematic design. For one thing, we don't
+                    // to see if the more precise rule is enabled. This gives the appearance of
+                    // false negatives in these low-level checks. This is a subtle topic
+                    // potentially arguing for redesign of the engine or our test expectations.
```
rwoll, line 66: agreed :) out-of-scope for this PR so I'll refrain from commenting further. You know my thoughts here :P. #Closed

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,10 +26,65 @@ public class WellKnownRegexPatternsTests
...
+                        pattern.Name == nameof(Unclassified16ByteHexadecimalString);
+
+                    result.Should().BeTrue(because: $"pattern '{pattern.GetType().Name}' should match '{example}'");
+
+                    if (moniker == null)
+                    {
+                        string matched = example.Substring(detection.Start, detection.End - detection.Start);
+                        moniker = pattern.GetMatchMoniker(matched);
+                        moniker.Should().NotBeNull(because: $"'{matched}' should produce a non-null moniker for {pattern.GetType().Name}' test data");
+                    }
```
rwoll, line 79: I'm having a very hard time understanding this test. Perhaps it's just Friday, but the cognitive load of this test is so high I cannot be confident of what it's testing or asserting.

I don't mean to bring it up again, but the combination of not seeing literal examples (i.e. only usage of `.GenerateTruePositives()`) and no literal named RuleIds/Patterns and **conditional/branching assertions** means there's *way too much* for my brain to process.

This is not intended to be a philisophical comment, but rather one that's simply stating: my brain is wired in such a way where I cannot process this test to properly review it.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: Out of scope per other comment.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: The test name `WellKnownRegexPatterns_GetMatchMonikerHardenedForOutOfOrderExecution` and the 9 line code comment in this test is intended to make clear what this test accomplishes, which is to aggressively call `GetMatchMoniker` as the first operation in processing all test data. This is the issue that was observed deploying our API in the wild and I therefore authored a test case mirroring this condition and provoked observed failures before attempting a product. I stand behind this test-driven approach and feel like I've invested significant in good code comments.

This is not to argue the point you've raised that we need investments in another test approach. But that is not the point of this PR and Nick has the baton for making this improvement.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: This test exercises 100% code coverage over the broken code paths observed in the wild. It also achieves 100% data coverage as all synthetic tests for all rules is passed through the previously broken code path. 

If others find the completeness of this change persuasive I would appreciate approval because deployment of 1.15 (which we need to burn) is blocking a significant partner.

We have identified two moving forward changes that need to be picked up as soon as our partner is unblocked, removing the iterations footgun and committing to more atomic tests.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
rwoll: Stamping *Approve* to unblock you. This looks low risk in introducing a regression, so my hesitation is not with the product code itself.

Despite the code comments (which I do appreciate!) — I'm struggling to understand this whole section:

```
                    bool result = 
                        detection != default ||
                        pattern.Name == nameof(Unclassified32CharacterString) ||
                        pattern.Name == nameof(Unclassified16ByteHexadecimalString);

                    result.Should().BeTrue(because: $"pattern '{pattern.GetType().Name}' should match '{example}'");

                    if (moniker == null)
                    {
                        string matched = example.Substring(detection.Start, detection.End - detection.Start);
                        moniker = pattern.GetMatchMoniker(matched);
                        moniker.Should().NotBeNull(because: $"'{matched}' should produce a non-null moniker for {pattern.GetType().Name}' test data");
                    }
```

Perhaps there's a way to re-write this bit of test code (read: not add more comments) for just this snippet so it's easier to understand. It may simply be variable naming or the like. Perhaps it's just my brain :)

rwoll (dismissed):
See comment, but otherwise LGTM.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
rwoll: Okie — looking at this a little longer, I think I've unwrapped a bit of the comrehensibility challenges:

Within the conditional assertion, the reader has to magically know there's something "special" about two rules:

`Unclassified32CharacterString`
`Unclassified16ByteHexadecimalString`

So consider structuring a bit differently to make it more clear:

```
if (pattern.Name == nameof(Unclassified32CharacterString) || pattern.Name == nameof(Unclassified16ByteHexadecimalString))
{
    detection.ShouldBeNull(because: "…")
}
else
{
     detection.ShouldNotBeNull(because: "…")
}
```

Then some well-considered `because`'s will help the reader unwrap all the work that's being done by that `bool result = …` line and the code comment above it.

If there's a concise word for what makes `Unclassified32CharacterString` and `Unclassified16ByteHexadecimalString` unique, you could even write it as:

```
var XXXXRules = [nameof(Unclassified32CharacterString), nameof(Unclassified16ByteHexadecimalString)];
if (XXXRules.Contains(pattern.Name)
{ … }
```

Yes, I'm ignoring the principle that tests should not have conditional assertions, but solely am doing so since this test already has it and hopefully my suggestion can make it easier to read.

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -995,7 +1000,18 @@ internal static byte[] ConvertFromBase64String(string text)
...
         text = TransformToStandardEncoding(text);
         text += RetrievePaddingForBase64EncodedText(text);
-        return Convert.FromBase64String(text);
+        byte[] result;
+
+        try
+        {
+            result = Convert.FromBase64String(text);
+        }
+        catch
```
nguerrera:
```suggestion
        catch (FormatException)
```

It's not as bad when general catch is scoped like this, but even here, swallowing ArgumentNullException could hide a bug.

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -995,7 +1000,18 @@ internal static byte[] ConvertFromBase64String(string text)
...
+        {
+            return null;
```
nguerrera: Please make this `bool TryConvertFromBase64String(string, out byte[])`. I'm not expecting something called ConvertXxx to return null and I wouldn't notice in PR if someone forgot to check.

nguerrera: It appears there's another caller that was not updated: GetBase64EncodedSignatureBytes.

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
michaelcfanning: That never should have been in there, sigh.

michaelcfanning: OK, turns out my final commit wasn't pushed. Since I didn't literally get the change out in a coherent form, I give up, I will go back and address the test comments. :)

nguerrera (approved):

Issue #143, opened 2025-02-25
chore: ensure generated patterns are checked in


rwoll:
The security-utilities build generate JSON files based on the RegexPatterns.

Ensure the build fails if the most up-to-date generated files are not checked in.

Currently, the build will silently succeed without these files being the most up to date.

nguerrera:
I'm embarrassed that I discovered this, filed #108, but never got around to fixing it.

PR #163, opened 2025-03-03
Various bug fixes.

michaelcfanning:
- BRK: Eliminate `SEC000/101.Unclassified32CharacterString` as noisy and not useful.
- BRK: Rename `SEC101/102.AdoPat` friendly name to `AdoLegacyPat`.
- BUG: Correct `SEC000/002.Unclassified16ByteHexadecimalString` id and rule name  on calling `GetMatchIdAndName` (where `SEC000/001.Unclassified64ByteBase64String` was returned incorrectly before).
- BUG: Resolve `System.FormatException: The input is not a valid Base-46 string` errors calling `SEC101/102.AdoPat.GetMatchIdAndName` by swallowing correct exception kind `ArgumentException` in `IsChecksumValid` helper.

I wish we could all be in a room together where I would invite everyone to hurl coffee cups, peanut shells, or anything else handy at my head. I completely failed to test all the code paths I strongly asserted I was covering in the last PR. Furthermore, Ross' confusion at a very strange set of conditions in a test I authored proved to be entirely warranted: these conditions indicate other bugs I simply wasn't seeing.

I am only human. `Festina lente`. I should have it tattooed in reverse on my forehead so that it's the first thing I see in the mirror every morning.


src/Microsoft.Security.Utilities.Core/PotentialSecurityKeys/SEC000_003_Unclassified32CharacterString.cs
```diff
@@ -1,102 +0,0 @@
...
-    {
-        public const string AzureContainerRegistryLegacyKey = nameof(AzureContainerRegistryLegacyKey);
-        public const string AadClientAppLegacyCredentials = nameof(AadClientAppLegacyCredentials);
-
-        /// <summary>
-        /// Detect 32-character Azure Active Directory client application legacy credentials.
-        /// The generated key is a 32-character string that contains alphanumeric characters
-        /// as well as symbols from the set: .=\-:[_@\*]+?
-        /// </summary>
-        public Unclassified32CharacterString()
```
michaelcfanning, line 19: >Unclassified32CharacterString [](http://example.com/codeflow?start=15&length=29)

Problem #1 in the strange test code paths Ross found suspicious: the test data generated by this rule was not in strict conformance across all its logic.

Basically, we have no general testing of the `UnclassifiedPotentialSecurityKeys` patterns list, which I could go slam into some other bucketing test. I am deferring that work for now.

For now, I'm ripping this rule out because it's noisy and not useful. The intent here was to capture an ancient variation of an generated AAD SPN secret, then the rule was renamed as a generic `Unclassified32CharacterString` because it fired too frequently and really it should have been deleted entirely. If there's an analysis here, it will argue its way back in.

src/Microsoft.Security.Utilities.Core/PotentialSecurityKeys/SEC000_002_Unclassified16ByteHexadecimalString.cs
```diff
@@ -22,7 +22,7 @@ public Unclassified16ByteHexadecimalString()
         DetectionMetadata = DetectionMetadata.HighEntropy | DetectionMetadata.Unclassified | DetectionMetadata.LowConfidence;
     }
 
-    public override Tuple<string, string>? GetMatchIdAndName(string match) => new Tuple<string, string>("SEC000/001", "Unclassified64ByteBase64String");
+    public override Tuple<string, string>? GetMatchIdAndName(string match) => new Tuple<string, string>("SEC000/002", nameof(Unclassified16ByteHexadecimalString));
```
michaelcfanning, line 25: >002 [](http://example.com/codeflow?start=112&length=3)

boo. cut and paste error. I added invariant tests that ensure no rules collide with each other in core metadata expression and also that `GetMatchIdAndName` results always match the declared rule id and name.

This is what we get for being 'cute' and allowing a single rule to be flexible in what ids it covers. That's useful but it tortures the design more than a little and leaves us vulnerable to bugs of this kind.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_102_AdoPat.cs
```diff
@@ -5,14 +5,14 @@
 
 namespace Microsoft.Security.Utilities
 {
-    public class AdoPat : RegexPattern
+    public class AdoLegacyPat : RegexPattern
```
michaelcfanning, line 8: >AdoLegacyPat [](http://example.com/codeflow?start=17&length=12)

This rule no longer applies to the latest ADO PAT format and thefore I'm renaming it.

src/Tests.Microsoft.Security.Utilities.Core/AdoLegacyPatTests.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Security.Utilities.Core
+{
+    [TestClass]
+    public class AdoLegacyPatTests
+    {
+        [TestMethod]
+        public void AdoLegacyPat_InvalidBase32Input()
```
michaelcfanning, line 12: >AdoLegacyPat_InvalidBase32Input [](http://example.com/codeflow?start=20&length=31)

Boo. For all my talk of test-driven development in the last PR, I ended up not shipping a correct fix! The reason? Late-breaking changes to a test I authored altered the code coverage and then at the very end a conversion from generated exception type (from `FormatException` to `ArgumentException`) left the code in a crashing condition,

Sigh. Even while waiting on a good unit test pattern, I should have started with simple, atomic tests to reveal bugs and prove they are fixed. Duly chastened, I return here with this review. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -57,26 +105,13 @@ public void WellKnownRegexPatterns_GetMatchMonikerHardenedForOutOfOrderExecution
                     string moniker = pattern.GetMatchMoniker(example);
 
                     var detection = masker.DetectSecrets(example).FirstOrDefault((d) => d.Id == pattern.Id);
-                    
-                    // Currently, some rules are tuned not to double-fire, i.e., they determine
```
nguerrera, line 61: Is there possibly enough fixed here to remove the `allowMultipleFindings=true` in another test. No need to block on this, but I am curious. #Resolved

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -85,7 +84,7 @@ public static IEnumerable<RegexPattern> HighConfidenceSecurityModelsIterator()
         new AzureCacheForRedisIdentifiableKey(),
         new AzureContainerRegistryIdentifiableKey(),
         new NuGetApiKey(),
-        new AdoPat(),                               // SEC101/102
+        new AdoLegacyPat(),                               // SEC101/102
```
nguerrera: nit: The comment is now misaligned with the others.
 #Resolved

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -271,5 +306,16 @@ public void WellKnownRegexPatterns_EnsureAllPatternsHaveCorrectCaptureGroups()
                 }
             }
         }
+
+        private static List<RegexPattern> GetAllPatterns()
```
nguerrera, line 312: I think there are other tests that could use this.

nguerrera (dismissed):

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
michaelcfanning: details matter thx

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: yes! that was just a bogus code add, ripped it out.

nguerrera (dismissed):

GeneratedRegexPatterns/LowConfidenceSecurityModels.json
```diff
@@ -6,13 +6,6 @@
     "Signatures": null,
     "DetectionMetadata": "LowConfidence"
   },
-  {
```
rwoll, line 9: Double check these files are all up-to-date. #143 is not yet implemented, so this comment serves as a reminder instead :P #Resolved

src/Tests.Microsoft.Security.Utilities.Core/AdoLegacyPatTests.cs
rwoll: > Sigh. Even while waiting on a good unit test pattern, I should have started with simple, atomic tests to reveal bugs and prove they are fixed. Duly chastened, I return here with this review.

❤️

src/Tests.Microsoft.Security.Utilities.Core/AdoLegacyPatTests.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            var classifier = new AdoLegacyPat();
+            string invalidInput = "=22222222222222222222222222";
+            var result = classifier.GetMatchIdAndName(invalidInput);
+            Assert.IsNull(result);
```
rwoll, line 17: Yay! This is a very easy to understand test <3.

For due diligence, if you have not already, run this test attached with a debugger and ensure the `null` is resulting from the expected error line getting hit. i.e. this is `null` for a particular reason, and not a general indeterminate reason.

Depending on perf constraints, there's more formal methods to handle this kinda testing — e.g. internal logging and asserting on specific logs are hit even if everything publicly is turned into a `null`. Out of scope for this PR, of course, but food for thought. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,6 +26,54 @@ public class WellKnownRegexPatternsTests
...
+        public void WellKnownRegexPatterns_AllRuleIdsAndNamesAreUnique()
+        {            
+            using var assertionScope = new AssertionScope();
+
+            var patterns = GetAllPatterns();
+
+            HashSet<string> ruleIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            HashSet<string> ruleNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            bool result;
```
rwoll: nit: why is result hoisted out of the loop? looks like it can just be inside it. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,6 +26,54 @@ public class WellKnownRegexPatternsTests
...
 
+        [TestMethod]
+        public void WellKnownRegexPatterns_AllRuleIdsAndNamesAreUnique()
+        {            
+            using var assertionScope = new AssertionScope();
+
+            var patterns = GetAllPatterns();
+
+            HashSet<string> ruleIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            HashSet<string> ruleNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
rwoll:
```suggestion
            HashSet<string> seenRuleIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
            HashSet<string> seenRuleNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

nit: these renames will make the loop code easier to read #Resolved

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,6 +26,54 @@ public class WellKnownRegexPatternsTests
...
+
+            bool result;
+            foreach (RegexPattern pattern in patterns)
+            {
+                result = ruleIds.Contains(pattern.Id);
```
rwoll: It doesn't look like the code is ever adding to `ruleIds` so I don't think this test is doing anything atm. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,6 +26,54 @@ public class WellKnownRegexPatternsTests
...
+                result.Should().BeFalse(because: $"Pattern '{pattern.GetType().Name}' should not share its Id with another rule: '{pattern.Id}'");
+
+                result = ruleNames.Contains(pattern.Name);
```
rwoll:
```suggestion
                bool duplicatePatternName = ruleNames.Contains(pattern.Name);
```

nit: instead of using generic `result` everywhere some more specific names will increase readability. #WontFix

rwoll (dismissed):
The invariant test for no duplicate ids looks like it's missing a key piece. See comments.

Otherwise, just some nits.

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
```diff
@@ -35,6 +37,16 @@ public static Regex GetOrCreateRegex(string pattern, RegexOptions options)
 #endif
         }
 
+        internal static string NormalizeGroupsPattern(string pattern)
+        {
+            if (pattern.IndexOf("?P<") != -1)
```
nguerrera: I'm sad to see this return because it showed up in profiling. 

In general, it's never great when allocations are done to look something up in a dictionary. 

At a bare minimum, we don't need two passes over the pattern for this normalization. string.Replace returns `this` when no match is found. The `if` is doubling the cost of the replacement.


 #Resolved

src/Microsoft.Security.Utilities.Benchmarks/RE2RegexEngine.cs
```diff
@@ -35,6 +31,28 @@ public IEnumerable<UniversalMatch> Matches(string input, string pattern, RegexOp
...
                 }
             }
+            else
+            {
+                if (Regex2.Matches(pattern, input, out List<Dictionary<string, FlexMatch>> matches, 256L * 1024L * 1024L))
+                {
+                    foreach (Dictionary<string, FlexMatch> match in matches)
+                    {
+                        FlexMatch flexMatch = match["0"];
+                        if (match.TryGetValue("refine", out FlexMatch refineMatch))
```
nguerrera: should this be `captureGroup` instead of hard-coded `"refine"`?

src/Microsoft.Security.Utilities.Benchmarks/RE2RegexEngine.cs
```diff
@@ -35,6 +31,28 @@ public IEnumerable<UniversalMatch> Matches(string input, string pattern, RegexOp
                     };
                 }
             }
+            else
+            {
+                if (Regex2.Matches(pattern, input, out List<Dictionary<string, FlexMatch>> matches, 256L * 1024L * 1024L))
```
nguerrera: Use a named arg or constant here. I was wondering if this is a max length, but it's max memory.

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
```diff
@@ -27,6 +27,8 @@ private CachedDotNetRegex()
 
         public static Regex GetOrCreateRegex(string pattern, RegexOptions options)
         {
+            pattern = NormalizeGroupsPattern(pattern);
```
nguerrera: Oh, here's an idea: move this into the construction of regex so it happens only once per pattern. Keep the denormalized pattern as the key.

`RegexCache.GetOrAdd(key, key => new Regex(NormalizeGroupsPattern(key.Pattern),...`
 #Pending

src/Microsoft.Security.Utilities.Benchmarks/RE2RegexEngine.cs
```diff
@@ -18,13 +18,9 @@ public class RE2RegexEngine : IRegexEngine
 #endif
```
nguerrera: I just noticed this is in benchmarks. Do tests use it? #Resolved

nguerrera: I see there are two copies. Can we use the same source in both benchmark and tests so that we are sure we're testing what we benchmark and vice versa.

docs/ReleaseHistory.md
```diff
@@ -11,11 +11,12 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
-# 1.16.0 - 03/03/2025
+# 1.16.0 - 03/04/2025
 - BRK: Eliminate `SEC000/101.Unclassified32CharacterString` as noisy and not useful.
 - BRK: Rename `SEC101/102.AdoPat` friendly name to `AdoLegacyPat`.
 - BUG: Correct `SEC000/002.Unclassified16ByteHexadecimalString` id and rule name  on calling `GetMatchIdAndName` (where `SEC000/001.Unclassified64ByteBase64String` was returned incorrectly before).
 - BUG: Resolve `System.FormatException: The input is not a valid Base-46 string` errors calling `SEC101/102.AdoPat.GetMatchIdAndName` by swallowing correct exception kind `ArgumentException` in `IsChecksumValid` helper.
+- BUG: `?P<name>` is now used throughout for named captures as this is required currently for RE2 compatibility. This reverts a clean-up in this area (that broke RE2) in 1.15.0.
```
nguerrera: RE2 was broken before this cleanup since most rules did not have the ?P before I touched anything.

nguerrera: This is more than a revert, in other words.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: nuts, thx

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: I don't do this. I set result to my test condition so that it's on one line. Then I follow it with an assertion with a strong readable message.

`result` isn't intended to encode something useful, the detailed, highly authored message it. 

i *could* allocate new variables with names but this tends to consume more space and i worry about readability.

let's punt for now i will think about your suggestion.

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
michaelcfanning: yes good catch will fix that and then let's take your suggestion.

src/Microsoft.Security.Utilities.Benchmarks/RE2RegexEngine.cs
michaelcfanning: have already done this, i should have warned, this isn't the current PR due to us having to accept your change first. i'm still reviewing and replying to comments here.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
rwoll: I read the code top to bottom. A well-named variable means I know how to read/check the line without the out-of-order flow.

Better yet if you want the `because` to be the sole communicator, skip using the variable altogether:

``` csharp
alreadySeeanRullName.Should().Not().Contains(pattern.Name, because: "…");
```

These are nits, so keep thinking on it.

GeneratedRegexPatterns/LowConfidenceSecurityModels.json
nguerrera: #143 is fixed now.

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
michaelcfanning: Fixed

src/Microsoft.Security.Utilities.Benchmarks/RE2RegexEngine.cs
michaelcfanning: Fixed. In my opinion, this argument is non-obvious. In some applications, code uses `secret` to indicate the refinement, in other contexts (like ours) that flag data other than secrets `refine` is used. `refinementCaptureGroup` would be better. doc comments would be a good idea as well.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
michaelcfanning: It is a good suggestion to encode more information about the test expectation in the variable name, I will keep this in mind moving forward.

docs/ReleaseHistory.md
nguerrera: Hiding my earlier messages on this and starting over since I was being personally defensive while being less than clear about my actual concern with the release note.

I believe this is true:

* 1.14.0 and earlier: All but a handful of patterns had RE2-*in*compatible syntax already. 
* 1.15.0: All patterns have RE2-*in*compatible syntax
* 1.16.0 (proposed): All patterns have RE2 compatible syntax.

If so, I don't think this release note quite captures the change.

src/Tests.Microsoft.Security.Utilities.Core/DetectionTests.cs
```diff
@@ -208,13 +208,13 @@ public void Detection_EqualComparisonUpdateWhenPropertiesChange()
             previousDetection = new Detection(currentDetection);
             Assert.AreEqual(currentDetection, previousDetection);
 
-            currentDetection.Start = Math.Max(1, (int)DateTime.UtcNow.Ticks % 99);
+            currentDetection.Start++;
```
nguerrera, line 211: Completely separate with zero action needed on this PR, but I am concerned about reference types that include mutable properties in their hash code.

See https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-9.0#notes-to-inheritors

This can cause objects to be lost in a dictionary.

src/Tests.Microsoft.Security.Utilities.Core/RE2RegexEngine.cs
```diff
@@ -36,6 +33,28 @@ public IEnumerable<UniversalMatch> Matches(string input, string pattern, RegexOp
                     };
                 }
             }
+            else
+            {
+                if (Regex2.Matches(pattern, input, out List<Dictionary<string, FlexMatch>> matches, 256L * 1024L * 1024L))
```
nguerrera: It would be nice to use a named argument or constant here. I was confused thinking this might be max input or match length, but it's max memory.

src/Tests.Microsoft.Security.Utilities.Core/WellKnownRegexPatternsTests.cs
```diff
@@ -26,22 +26,74 @@ public class WellKnownRegexPatternsTests
...
 
             var patterns = new List<RegexPattern>();
-
+            
             patterns.AddRange(WellKnownRegexPatterns.DataClassification);
             patterns.AddRange(WellKnownRegexPatterns.PreciselyClassifiedSecurityKeys);
-            patterns.AddRange(WellKnownRegexPatterns.UnclassifiedPotentialSecurityKeys);
+            patterns.AddRange(WellKnownRegexPatterns.UnclassifiedPotentialSecurityKeys);            
+
+            patterns.Add(new AzureFunctionIdentifiableKey());
```
nguerrera: What is happening here that this rule needs to be added one-off. Should this be using GetAllPatterns()?

nguerrera (dismissed):

src/Tests.Microsoft.Security.Utilities.Core/RE2RegexEngine.cs
nguerrera: Hmm, seems there was a race in my review and this getting fixed.

src/Tests.Microsoft.Security.Utilities.Core/DetectionTests.cs
michaelcfanning: As mentioned offline, these tests and this code condition inherited from the ADO pipeline agent `ReplacementPosition` class, which was elaborated on in our use to represent a detection, consumable in non-masking cases. My suggestion is that we go update that code to consume our package, and then we can work on any internals where we see improvements to make. https://github.com/microsoft/security-utilities/issues/169

nguerrera (approved):

PR #165, opened 2025-03-04
Fixes for upcoming 1.16.0 release

michaelcfanning:
- BRK: Eliminate `SEC000/101.Unclassified32CharacterString` as noisy and not useful.
- BRK: Rename `SEC101/102.AdoPat` friendly name to `AdoLegacyPat`.
- BUG: Correct `SEC000/002.Unclassified16ByteHexadecimalString` id and rule name  on calling `GetMatchIdAndName` (where `SEC000/001.Unclassified64ByteBase64String` was returned incorrectly before).
- BUG: Resolve `System.FormatException: The input is not a valid Base-46 string` errors calling `SEC101/102.AdoPat.GetMatchIdAndName` by swallowing correct exception kind `ArgumentException` in `IsChecksumValid` helper.
- BUG: `?P<name>` is now used throughout for named captures as this is required currently for RE2 compatibility. This reverts a clean-up in this area (that broke RE2) in 1.15.0.


src/Microsoft.Security.Utilities.Benchmarks/Microsoft.Security.Utilities.Benchmarks.csproj
```diff
@@ -14,6 +14,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Security.Utilities.Core\Microsoft.Security.Utilities.Core.csproj" />
+    <ProjectReference Include="..\Tests.Microsoft.Security.Utilities.Core\Tests.Microsoft.Security.Utilities.Core.csproj" />
```
michaelcfanning, line 19: >Include [](http://example.com/codeflow?start=22&length=7)

This dependency allows us to remove a copy of the RE2 engine wrapper.

nguerrera:
Back into main from where? Is this already in a release branch? #Closed

michaelcfanning:
Yes! We have a `release/v1.16.0` branch. and so I have to merge that branch back into main. this branch was forked before your large change.

---
In reply to: [2699034783](https://github.com/microsoft/security-utilities/pull/165#issuecomment-2699034783) [](http://example.com/codeflow?ancestors=2699034783)

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -157,13 +157,14 @@ public void SecretMasker_LowConfidenceSecurityModels_Masking()
     private void ValidateSecurityModelsMasking(IEnumerable<RegexPattern> patterns, IRegexEngine engine, bool lowEntropyModels)
 
     {
+        using var assertionScope = new AssertionScope();
```
michaelcfanning, line 160: >assertionScope [](http://example.com/codeflow?start=18&length=14)

this actually has no effect. i've been unable to figure out the magic to allow all failures in this test to be accumulated and reported. perhaps this logic would have to be inlined i the upstream caller? putting a pin in it for now

rwoll (commented):
Where did `1-16-0-back-to-main` come from?

I would expect the sequence of events to be:

1. Land all new features and fixes on `main` (which were the smaller PRs).
2. Then cut a branch from `main` for the `v1.16.0` release.

If something misses the cut point, then it still goes onto main and then you open a PR cherry-pick over of that main commit to release branch.

rwoll:
What happened to https://github.com/microsoft/security-utilities/pull/163 PR? That was almost fully reviewed and would be the fixes, I think. Then v1.16.0 can be cut. If it missed the original cut from main, the standard flow would be to land the fix on main, then cherry-pick the fix from main onto the release branch.

src/Tests.Microsoft.Security.Utilities.Core/DetectionTests.cs
```diff
@@ -208,13 +208,13 @@ public void Detection_EqualComparisonUpdateWhenPropertiesChange()
             previousDetection = new Detection(currentDetection);
             Assert.AreEqual(currentDetection, previousDetection);
 
-            currentDetection.Start = Math.Max(1, (int)DateTime.UtcNow.Ticks % 99);
+            currentDetection.Start++;
```
michaelcfanning, line 211: >Start [](http://example.com/codeflow?start=29&length=5)

There have been several poorly thought out attempts to make this test not occasionally fail due to randomly matching data fixed earlier in execution. This is the latest poorly thought out attempt.

I think it's right this time? Basically we constrain all data for these properties in a range from 0-98 and for all values it is both safe to increment them by one and by doing so we should invalidate equality and pass the test.

nguerrera:
> If something misses the cut point, then it still goes onto main and then you open a PR cherry-pick over of that main commit to release branch.

I gave different advice yesterday, sorry. I was also willing to accept a PR to the release branch first followed by port back to main. Mind you, I only offered this after #163 was closed and I didn't understand the idea that this was coming back from a release branch, which should not have been the case since the changes were not approved yet. It wasn't actually the case, but I felt that seeing an actual PR into a release branch would make me more confident of what's already in and what's going in. 

Over the years, I've seen many different approaches, and I think there's only one thing that is critical to me: 

`All commits to release branches (cherry-pick or otherwise) must be reviewed, signed off, and pass PR validation.`

Regardless of the order of operations, it appears we have some technical debt in this repo for making a release that has any divergence from main for a hotfix:

1. release/* branches were not protected. I fixed this.
2. PR validation does not run on release branches: I will send a PR. (Edit: Done. #168)

nguerrera:
I'm changing the PR title to reflect that these changes are not in fact coming back from anywhere. I will review this and #163 side-by-side and port comments that were unresolved, and also ensure that I have reviewed any delta since #163.

**EDIT**: This is proving too complicated. We should close this, reopen #163 and add any additional changes to review to #163.

nguerrera:
> Yes! We have a release/v1.16.0 branch. and so I have to merge that branch back into main. this branch was forked before your large change.

We did not in fact have a release/v1.16.0 branch at the time of that comment.  There is one now, but it is identical to release/v1.15.0 at this time.

src/Tests.Microsoft.Security.Utilities.Core/DetectionTests.cs
nguerrera: I wanted to understand what this test is doing to review this. It points to a larger design issue with mutable properties of a reference type factoring in GetHashCode()/Equals(). This is a footgun that can cause objects to be lost in dictionaries.

See https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-9.0#notes-to-inheritors

Why is Detection mutable?

nguerrera: #169

PR #176, opened 2025-03-28
Sort properties in exported JSON

nguerrera:
Currently, property order is influenced by whether or not a property is overridden or introduced by a derived class. This change orders the properties by name irrespective of where they appear in the class hierarchy.

Submitting it on its own so that a pending change that removes several overrides will be easier to review.

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -11,7 +11,10 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
-# 1.16.0 - 03/05/2025
+# UNRELEASED
+- NEW: Sort properties by name in GeneratedRegexPatterns/*.json.
+
+# 1.16.0 - 03/05/2025/
 - BRK: Eliminate `SEC000/101.Unclassified32CharacterString` as noisy and not useful.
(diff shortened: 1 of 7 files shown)


docs/ReleaseHistory.md
```diff
@@ -11,7 +11,10 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
-# 1.16.0 - 03/05/2025
+# UNRELEASED
+- NEW: Sort properties by name in GeneratedRegexPatterns/*.json.
+
+# 1.16.0 - 03/05/2025/
```
```suggestion
# 1.16.0 - 03/05/2025
```

michaelcfanning (approved):
:shipit:

docs/ReleaseHistory.md
```diff
@@ -11,6 +11,9 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- NEW: Sort properties by name in GeneratedRegexPatterns/*.json.
```
michaelcfanning, line 15: I'm on the fence regarding whether this change is worth documenting at all.  Or whether it's a `NEW` label. We might have labeled this as clean-up, fundamentals, work in a separate project. `FUN`. SEV4, honestly, just leave it alone.

michaelcfanning (approved):

PR #177, opened 2025-03-28
Remove expensive re-computation in property getters

### LINKED ISSUE #145: Avoid expensive re-computation in property getters
A property get should be on the order of a field read in cost. Lazy initialization is OK, but allocating or otherwise expensive re-computation on each and every access is not. After all, it looks just like a field read syntactically, so nobody looks at a line of code like this and realizes that `Pattern` will do work and allocate:

https://github.com/microsoft/security-utilities/blob/29626a28388b862ba051f7efc320002195b9b730/src/Microsoft.Security.Utilities.Core/RegexPattern.cs#L195

But then we have property getters like this:

https://github.com/microsoft/security-utilities/blob/29626a28388b862ba051f7efc320002195b9b730/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/Azure32ByteIdentifiableKey.cs#L16

Which chain to more properties like this:

https://github.com/microsoft/security-utilities/blob/29626a28388b862ba051f7efc320002195b9b730/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/IdentifiableKey.cs#L20

And then this:

https://github.com/microsoft/security-utilities/blob/4de8259b8901c5e652a937ba9622173440cce8a8/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_178_AzureIotHubIdentifiableKey.cs#L16

I suggest that we scrub the whole RegexPattern hierarchy such that all of the constant properties are computed once during class construction. It is non-trivial to fix these one at a time as there are ordering issues that result due to the interdependencies between these properties.

nguerrera:
Fix #145

### DIFF
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -21,7 +21,7 @@
 - NEW: Sort properties by name in GeneratedRegexPatterns/*.json.
 - PRF: Remove unnecessary and expensive recomputation of `RegexPatter.Pattern`, `RegexPattern.Signatures`, and `IdentifiableKey.ChecksumSeeds` on every property access.
 
-# 1.16.0 - 03/05/2025/
+# 1.16.0 - 03/05/2025
 - BRK: Eliminate `SEC000/101.Unclassified32CharacterString` as noisy and not useful.
 - BRK: Rename `SEC101/102.AdoPat` friendly name to `AdoLegacyPat`.
 - BRK: `IdentifiableScan` no longer supports stream input. The following API are removed. Use `IdentifiableScan.DetectSecrets(string)`.
--- a/docs/ReleaseHistory.md
+++ b/docs/ReleaseHistory.md
@@ -12,9 +12,16 @@
 - FNS => False negative reduction in static analysis.
 
 # UNRELEASED
+- BRK: `RegexPattern.Pattern` is no longer virtual.
+- BRK: `RegexPattern.Signatures` is no longer virtual.
+- BRK: `IdentifiableKey.ChecksumSeeds` is no longer virtual.
+- BRK: `RegexPattern.RotationPeriod` is no longer publicly settable.
+- BRK: `IdentifiableKey.RegexNormalizedSignature` is removed.
+- BRK: Abstract classes `IdentifiableKey`, `Azure32ByteIdentifiableKey`, `Azure64ByteIdentifiableKey`, and `AzureMessagingIdentifiableKey` now require derived classes to pass their signature to the base constructor.
 - NEW: Sort properties by name in GeneratedRegexPatterns/*.json.
+- PRF: Remove unnecessary and expensive recomputation of `RegexPatter.Pattern`, `RegexPattern.Signatures`, and `IdentifiableKey.ChecksumSeeds` on every property access.
 
-# 1.16.0 - 03/05/2025
+# 1.16.0 - 03/05/2025/
 - BRK: Eliminate `SEC000/101.Unclassified32CharacterString` as noisy and not useful.
 - BRK: Rename `SEC101/102.AdoPat` friendly name to `AdoLegacyPat`.
 - BRK: `IdentifiableScan` no longer supports stream input. The following API are removed. Use `IdentifiableScan.DetectSecrets(string)`.
--- a/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/IdentifiableKey.cs
+++ b/src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/IdentifiableKey.cs
@@ -13,22 +13,22 @@ namespace Microsoft.Security.Utilities
 {
     public abstract class IdentifiableKey : RegexPattern, IIdentifiableKey, IHighPerformanceScannableKey
     {
-        public IdentifiableKey()
+        protected IdentifiableKey(string signature)
         {
             RotationPeriod = TimeSpan.FromDays(365 * 2);
             DetectionMetadata = DetectionMetadata.Identifiable;
+            Signatures = signature.ToSet();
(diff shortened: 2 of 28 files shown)


src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/IdentifiableKey.cs
```diff
@@ -13,22 +13,22 @@ namespace Microsoft.Security.Utilities
...
             RotationPeriod = TimeSpan.FromDays(365 * 2);
             DetectionMetadata = DetectionMetadata.Identifiable;
+            Signatures = signature.ToSet();
+            ChecksumSeeds = Array.Empty<ulong>();
         }
 
         // Identifiable key patterns must have exactly one signature.
         internal string Signature => Signatures!.Single();
 
-        public string RegexNormalizedSignature => Regex.Escape(Signature);
```
nguerrera, line 25: On a prior PR of mine, @rwoll asked for this to be renamed, but I left it since it was pre-existing. 

Looking at it again now, given that:
1. Its expense was non-obvious (main point of this PR)
2. We don't need to store it long-term
3. The contentious name is longer than the implementation

I've just inlined `Regex.Escape` when needed. #Resolved

docs/ReleaseHistory.md
```diff
@@ -11,7 +11,17 @@
...
-# 1.16.0 - 03/05/2025
+# UNRELEASED
+- BRK: `RegexPattern.Pattern` is no longer virtual.
+- BRK: `RegexPattern.Signatures` is no longer virtual.
+- BRK: `IdentifiableKey.ChecksumSeeds` is no longer virtual.
+- BRK: `RegexPattern.RotationPeriod` is no longer publicly settable.
+- BRK: `IdentifiableKey.RegexNormalizedSignature` is removed.
+- BRK: Abstract classes `IdentifiableKey`, `Azure32ByteIdentifiableKey`, `Azure64ByteIdentifiableKey`, and `AzureMessagingIdentifiableKey` now require derived classes to pass their signature to the base constructor.
+- PRF: Remove unnecessary and expensive recomputation of `RegexPatter.Pattern`, `RegexPattern.Signatures`, and `IdentifiableKey.ChecksumSeeds` on every property access.
+- NEW: Sort properties by name in GeneratedRegexPatterns/*.json.
```
michaelcfanning, line 21: >NEW [](http://example.com/codeflow?start=2&length=3)

NIT: `NEW` labeled items should appear before `PRF`. we group and order release note entries following the ordering in the `Definitions` legend.

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
```diff
@@ -1,51 +1,51 @@
...
-    ],
-    "DetectionMetadata": "ObsoleteFormat, HighEntropy, MediumConfidence"
+    ]
   },
   {
-    "Pattern": "(?:^|[^0-9A-Za-z-_.])e[0-9A-Za-z-_=]{23,}\\.e[0-9A-Za-z-_=]{23,}\\.[0-9A-Za-z-_=]{24,}(?:[^0-9A-Za-z-_]|$)",
+    "DetectionMetadata": "HighEntropy, MediumConfidence",
     "Id": "SEC101/528",
     "Name": "GenericJwt",
+    "Pattern": "(?:^|[^0-9A-Za-z-_.])e[0-9A-Za-z-_=]{23,}\\.e[0-9A-Za-z-_=]{23,}\\.[0-9A-Za-z-_=]{24,}(?:[^0-9A-Za-z-_]|$)",
```
michaelcfanning: >- [](http://example.com/codeflow?start=116&length=1)

unescaped hyphens make me nervous in these range constructs. if someone introduced a character before this hyphen it would change it to a separator between a start and trailing char. if you add an underscore before the hyphen, for example, the hyphen would now be allowed to match.

you didn't create the problem, but can we add a work item to ensure this hyphen and all other hyphens in ranges are escaped?

michaelcfanning (dismissed):
:shipit:

docs/ReleaseHistory.md
nguerrera: Fixed.

nguerrera:
@michaelcfanning Release notes fixed, squashed over main that now includes #176. Needs restamp. 

Edit: Also filed #178 for pre-existing issue with hyphens.

GeneratedRegexPatterns/MediumConfidenceSecurityModels.json
nguerrera: Filed #178

michaelcfanning (dismissed):
:shipit:

docs/ReleaseHistory.md
```diff
@@ -12,9 +12,16 @@
...
+- BRK: `RegexPattern.Signatures` is no longer virtual.
+- BRK: `IdentifiableKey.ChecksumSeeds` is no longer virtual.
+- BRK: `RegexPattern.RotationPeriod` is no longer publicly settable.
+- BRK: `IdentifiableKey.RegexNormalizedSignature` is removed.
+- BRK: Abstract classes `IdentifiableKey`, `Azure32ByteIdentifiableKey`, `Azure64ByteIdentifiableKey`, and `AzureMessagingIdentifiableKey` now require derived classes to pass their signature to the base constructor.
 - NEW: Sort properties by name in GeneratedRegexPatterns/*.json.
+- PRF: Remove unnecessary and expensive recomputation of `RegexPatter.Pattern`, `RegexPattern.Signatures`, and `IdentifiableKey.ChecksumSeeds` on every property access.
 
-# 1.16.0 - 03/05/2025
+# 1.16.0 - 03/05/2025/
```
```suggestion
# 1.16.0 - 03/05/2025
```

Oops.

michaelcfanning (approved):

Issue #169, opened 2025-03-05
Review code for mutable reference types that may get lost in dictionaries


michaelcfanning:
E.g., our `Detection` class. 

https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-9.0#notes-to-inheritors

nguerrera:
Detection should be immutable and it shouldn't need a copy constructor if so.

nguerrera:
I've reviewed and not found any other types that Detection, which is fixed by #190.

PR #179, opened 2025-04-01
Readable user-facing secret 'labels'.

### LINKED ISSUE #173: Add readable secret kind label and user-facing string construction.
We will need to port the truncation helper for this.

michaelcfanning:
- NEW: Add `RegexPattern.Label` and `Detection.Label` properties that comprise a readable description of a secret kind, e.g., `an Azure Functions access key`.
- NEW: Add `Detection.FormattedMessage` extension method that returns user-facing strings such as `'…444444' is a NuGet API key. The correlating id for this detection is 3wsjgo3hBjSPHdIc2Td1.`.

A label in this use indicates a fragment of descriptive text suitable for inserting in a user-facing string or UI control, e.g., `an Azure DevOps personal access token (PAT)`. This data is distinct from the rule 'name' which comprises a pascal-cased readable name for the detection kind, e.g. `AdoPat`.

As usual, there is an unexpected and unwanted nuance in that various identifiers require either `a` or `an` as a leading prefix.

This change does not formalize any localization strategy for this content. We should discuss with a partner team how to make this work for their scenarios.

Resolves #173.


src/Microsoft.Security.Utilities.Cli/ScanCommand.cs
```diff
@@ -49,13 +50,14 @@ internal int ProcessInputFile(ScanOptions options)
...
+                string fileName = Path.GetFileName(path);
+                string contents = File.ReadAllText(path);
                 bool foundAtLeastOne = false;
 
-                foreach (var detection in scan.DetectSecrets(file))
+                foreach (var detection in scan.DetectSecrets(contents))
                 {
                     foundAtLeastOne = true;
-                    Console.WriteLine("Found {0} ('{1}') at position {2}", detection.Id, detection.RedactionToken, detection.Start + detection.Length);
+                    Console.WriteLine($"{fileName} ({detection.Start},{detection.End}): {detection.Moniker} : {detection.FormattedMessage(contents)}");
```
michaelcfanning: >WriteLine [](http://example.com/codeflow?start=28&length=9)

Generates output such as the following:

> AllExamplesAllSecretKinds.txt (103250,103290): SEC101/050.NpmAuthorKey : '…mMjs5P' is an NPM author key. The correlating id for this detection is Yfeb9REnWVeBQQ7hJ7iN.

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
```diff
@@ -16,5 +16,62 @@ public static ISet<string> ToSet(this IEnumerable<string> value)
...
+        /// of the input string is removed.
+        /// </summary>
+        /// <param name="text">The text to truncate.</param>
+        /// <param name="lengthExclusiveOfEllipsis">The desired length of the truncated text, exclusive of the ellipsis, if added.</param>
+        /// <returns>The rightmost truncated contents of the strength or the entire string (if its length is equal to or less than the specified length).</returns>
+        public static string Truncate(this string text, int lengthExclusiveOfEllipsis = 6)
+        {
+            text ??= string.Empty;
+            string truncatedText = text.TrimEnd('=');
+            string suffix = new string('=', text.Length - truncatedText.Length);
```
nguerrera: The handling of `=` suffix isn't documented.

Is it needed? It's making the whole method implementation hard to understand. #Resolved

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
```diff
@@ -16,5 +16,62 @@ public static ISet<string> ToSet(this IEnumerable<string> value)
...
+
+            if (truncatedText.Length <= lengthExclusiveOfEllipsis)
+            {
+                return text;
```
nguerrera: My gut reaction is shouldn't this be if `text.Length <= lengthExclusiveOfEllipsis`.

If I'm reading right (I'm not sure I am) `Truncate("Foo====================================================")` will not truncate even though it's longer than lengthExclusiveOfEllipsis?

 #Resolved

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
```diff
@@ -16,5 +16,62 @@ public static ISet<string> ToSet(this IEnumerable<string> value)
...
+        }
+
+        /// <summary>
+        /// Truncates a string to a specified length, adding an ellipsis if any
+        /// of the input string is removed.
+        /// </summary>
+        /// <param name="text">The text to truncate.</param>
+        /// <param name="lengthExclusiveOfEllipsis">The desired length of the truncated text, exclusive of the ellipsis, if added.</param>
+        /// <returns>The rightmost truncated contents of the strength or the entire string (if its length is equal to or less than the specified length).</returns>
+        public static string Truncate(this string text, int lengthExclusiveOfEllipsis = 6)
```
nguerrera: `this string` 😔

I almost got more mad but then saw the declaring type is internal. I would put a hard no on shipping any public extension methods to primitives.

I don't like internal extension methods on primitive types either. It's so confusing reading code that calls random stuff on primitives. #Resolved

nguerrera: Also, the special casing of `=` makes this even more questionable because "Truncate" looks like something general-purpose that could have actually been on string itself, but then has some random opinions.

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
```diff
@@ -16,5 +16,62 @@ public static ISet<string> ToSet(this IEnumerable<string> value)
...
+            }
+
+            truncatedText = truncatedText.Substring(truncatedText.Length - lengthExclusiveOfEllipsis);
+
+            bool charsElided = truncatedText.Length != text.Length;
+
+            // "\u2026" == "…"
+            return (charsElided ? "\u2026" : string.Empty) +
+                   truncatedText.Substring(truncatedText.Length - lengthExclusiveOfEllipsis) +
+                   (charsElided ? suffix : string.Empty);
```
nguerrera: OK, here we're adding the suffix of `=` signs back, so I guess the goal is to include all trailing `=` always?

Maybe this is so you get at least lengthExclusiveOfEllipsis non-equals signs and maybe we're only expecting = on the right to be base64 padding, so not expecting arbitrarily many?

I think we should either remove the `=` special case, or document it and make it more defensive against lots of `=`. #Resolved

src/Tests.Microsoft.Security.Utilities.Core/CoreExtensionMethodsTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        [TestMethod]
+        public void CoreExtensionMethod_FormattedMessage()
+        {
+            var classifier = new SecretScanningSampleToken();
+            var masker = new SecretMasker([classifier]);
+            string example = classifier.GenerateTruePositiveExamples().First();
+            var detections = classifier.GetDetections(example, generateCrossCompanyCorrelatingIds: true);
+            detections?.Count().Should().Be(1);
+            var detection = detections.First();
+            string truncated = example.Truncate();
```
nguerrera: We're allowing the implementation of truncate to decide what truncation is expected and I don't see any other tests of truncate.

Please add truncate coverage and include the = special case if we keep it. #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_UnclassifiedLegacyCommonAnnotatedSecurityKey.cs
```diff
@@ -1,80 +1,81 @@
...
-            "=",
-        };
-
-        private static readonly string[] s_nonInvalidatingSuffixes = new[]
-        {
-            ";",
-        };
-    }
-}
+﻿// Copyright (c) Microsoft. All rights reserved.
```
nguerrera, line 1: Is there some whitespace diff I can't see? GitHub thinks every line is changed without ignore whitespace.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -144,39 +144,20 @@ public string MaskSecrets(string input)
...
-                    new Detection(detection.Id,
-                                  detection.Name,
-                                  detection.Start,
-                                  detection.Length,
-                                  detection.Metadata,
-                                  detection.RotationPeriod,
-                                  detection.CrossCompanyCorrelatingId,
-                                  detection.RedactionToken);
-
+                currentDetection = new Detection(detection);
```
nguerrera, line 147: Is the copying here because Detection is mutable? Related: #169. It would be nice to make it immutable and then we wouldn't need to make extra instances. Not for this PR. #Pending

src/Microsoft.Security.Utilities.Cli/ScanCommand.cs
```diff
@@ -2,6 +2,8 @@
 
 #nullable disable
 
+using Microsoft.VisualBasic;
+
```
nguerrera: ? #Resolved

src/Microsoft.Security.Utilities.Cli/ScanCommand.cs
```diff
@@ -10,6 +12,7 @@ public class ScanCommand
     {
         public ScanCommand()
         {
+            Console.OutputEncoding = System.Text.Encoding.UTF8;
```
nguerrera: I suggest moving this to Main. #Resolved

michaelcfanning: yes, good note

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
michaelcfanning: See code moved to `Detections` static helper. I've documented the handling around the padding and also put limits on how/when we special-case this data. I really went back and forth on whether to require the caller to handle it.

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
michaelcfanning: fixed, in a non-performant way, see code in `Detections`. now, we attempt to decode the string as base64 and only in this case do we elide padding (i.e., on valid base64 data).

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
michaelcfanning: fixed see code in `Detections` class.

src/Microsoft.Security.Utilities.Core/CoreExtensionMethods.cs
michaelcfanning: made it defensive, thanks for the note. see `Detections`

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_200_UnclassifiedLegacyCommonAnnotatedSecurityKey.cs
michaelcfanning: apparently the addition of label is causing this craziness. i literally copied over the file again and re-added the label code to same effect. there could still be some underlying whitespace difference that was getting provoked by my copy gesture. i could try to chase this down in Notepad++ or something but am not sure it's worth it.

nguerrera: OK. Weird, but I agree not worth it.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
michaelcfanning: um, looking at this, i'm not sure why these copies are necessary. i just reflexively switched to the alternate ctor to minimize vertical space. let's fix this up in #169.

src/Microsoft.Security.Utilities.Core/Detections.cs
```diff
@@ -0,0 +1,90 @@
...
+                // retained in the truncated strength does not apply to the 
+                // length constraint. 
+                Convert.FromBase64String(truncatedText);
+                truncatedText = text.TrimEnd('=');
+            }
+            catch (FormatException)
+            {
+            }
+
+            string suffix = new string('=', text.Length - truncatedText.Length);
```
nguerrera, line 73: This first-chance exception might get annoying when debugging. I think we could do what the new description literally says and remove up to two equals regardless of valid base64 or not.

Something like this?

```C#
string suffix = string.Empty;
if (text.EndsWith("==", System.StringComparison.Ordinal))
{
    suffix = "==";
}
else if (text.EndsWith('='))
{
    suffix = "=";
}

string truncatedText = text.Substring(0, text.Length - suffix.Length);
```

nguerrera (dismissed):

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Added a comment to #169

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -312,13 +315,14 @@ public static string GenerateSha256Hash(string text)
...
 
-    [DataMember(Order = 4)]
+    [DataMember(Order = 5)]
     public string Pattern { get; protected set; }
 
     /// <summary>
     /// Gets or sets an opaque, stable identifier for the pattern (corresponding to a SARIF 'reportingDescriptorReference.id' value).
     /// </summary>
     [DataMember(Order = 1)]
+
```
nguerrera: This blank line is really good at coming back, lol. #Resolved

michaelcfanning: nuts fixed

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
```diff
@@ -11,6 +12,7 @@
   {
     "Id": "SEC101/156",
     "Name": "AadClientAppIdentifiableCredentials",
+    "Label": "an Azure Entra client app password",
```
nguerrera, line 15: Ignore this, testing if VS Code in browser works on tis repo.

nguerrera: Woohoo, it does!

nguerrera (dismissed):

src/Microsoft.Security.Utilities.Core/Detections.cs
michaelcfanning: I'd like to ship what we have. Here's why, I actually like the fact that we only attempt this operation for valid base64 strings. If I took your suggestion a bunch of thoughtful tests would begin to fail. :)

What I suggest is that I go back in later and add a more developed static 'IsValidBase64' check (rather than inferring this from the fact that Convert.FromBase64String doesn't raise an exception. #187

nguerrera: Sounds good.

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
michaelcfanning: Amazing. Awesome.

nguerrera (approved):

Issue #153, opened 2025-02-28
Update dotnet format settings and enforcement


nguerrera:
#132 removed stylecop analyzers because they weren't working on newer C# constructs. This could probably have been resolved with a new stylecop version, but we should use `dotnet format` instead like microsoft/cask.

There is an existing `dotnet format` step in CI but it is either not adequately configured to enforce our style preferences or not working.

PR #180, opened 2025-04-01
Make ordering of regex properties deterministic.

michaelcfanning:
Previously, we simply alphabetized the Regex properties to make their order deterministic. This change preserves determinism by explicitly specifying `[DataMember(Order = X)]` (metadata that JSON.NET honors when serializing).

Now, we can consciously order the properties for readability while ensuring they remain in this constant order. I have accordingly moved the detections metadata information to follow the id and readable name (which in my opinion are most helpful to review early).

Finally, this change elides additional properties in derived types that are specific to scenarios such as identifiable secret detection. These additional properties are no universally applicable to regexes and in most cases relate to sample generation. It is not clear this metadata is useful to JSON file consumers due to the inconsistent and specialized nature of this data.

### DIFF
--- a/src/Microsoft.Security.Utilities.Core/Base62.cs
+++ b/src/Microsoft.Security.Utilities.Core/Base62.cs
@@ -15,7 +15,7 @@ namespace Base62
     {
         [ThreadStatic]
         private static StringBuilder s_sb;
-        
+
         private const string DefaultCharacterSet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
         private const string InvertedCharacterSet = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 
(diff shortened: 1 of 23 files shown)


src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -33,11 +35,11 @@ public RegexPattern(string id,
 
         Id = id;
         Name = name;
-        m_regexOptions = regexOptions;
-        Signatures = signatures;
+        DetectionMetadata = patternMetadata;
         RotationPeriod = rotationPeriod;
+        Signatures = signatures;
+        m_regexOptions = regexOptions;
```
michaelcfanning, line 41: >m_regexOptions [](http://example.com/codeflow?start=8&length=14)

this assignment ordering now mirrors parameter ordering.

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -315,13 +318,17 @@ public static string GenerateSha256Hash(string text)
     /// <summary>
     /// Gets or sets an opaque, stable identifier for the pattern (corresponding to a SARIF 'reportingDescriptorReference.id' value).
     /// </summary>
+    [DataMember(Order = 1)]
+
```
nguerrera: Extra line of whitespace here.

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -403,4 +412,4 @@ private protected static string MakeHighPerformancePattern(string pattern, strin
         return "^" + pattern;
     }
 #endif
-}
+}
```
nguerrera, line 414: You can leave this, but just noting that doing #153 would prevent final-newline inconsistency from happening.

nguerrera (dismissed):

src/Microsoft.Security.Utilities.Core/Base62.cs
```diff
@@ -118,30 +118,30 @@ public static T FromBase62<T>(this string base62, bool inverted = false)
             switch (Type.GetTypeCode(typeof(T)))
             {
                 case TypeCode.String:
-                return (T)Convert.ChangeType(Encoding.UTF8.GetString(array), typeof(T), CultureInfo.InvariantCulture);
+                    return (T)Convert.ChangeType(Encoding.UTF8.GetString(array), typeof(T), CultureInfo.InvariantCulture);
```
nguerrera, line 121: There's a whole bunch of reformatting like this coming in with the last commit. Can you revert that commit and just fix the extra newline?

I will do #153 soon and fix formatting like this in a separate PR before turning on the enforcement.

PR #181, opened 2025-04-01
Flow all regex pattern properties from regex pattern to `IdentifiableScan` detection instances.

michaelcfanning:
- BUG: `IdentifiableScan` now properly flows `RegexPattern.RotationPeriod` to `Detection` instances. `Detection.RotationPeriod` previously always retained `default` as a value.
- BUG: `IdentifiableScan` now properly flows `RegexPattern.DetectionMetadata` to `Detection` instances. `Detection.DetectionMetadata` was previously hard-coded as `DetectionMetadata.HighEntropy`.

### DIFF
--- a/src/Tests.Microsoft.Security.Utilities.Core/IdentifiableScanTests.cs
+++ b/src/Tests.Microsoft.Security.Utilities.Core/IdentifiableScanTests.cs
@@ -22,8 +22,9 @@ namespace Tests.Microsoft.Security.Utilities.Core
             var cask = new CommonAnnotatedSecurityKey();
             var examples = cask.GenerateTruePositiveExamples().ToList();
 
-            var masker = new IdentifiableScan(WellKnownRegexPatterns.HighConfidenceMicrosoftSecurityModels,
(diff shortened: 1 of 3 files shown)


src/Tests.Microsoft.Security.Utilities.Core/IdentifiableScanTests.cs
```diff
@@ -86,5 +87,33 @@ public void IdentifiableScan_IdentifiableKeys()
                 }
             }
         }
+
+        [TestMethod]
+        public void IdentifiableScan_BehaviorMatchesStandardScanner()
```
michaelcfanning, line 92: >IdentifiableScan_BehaviorMatchesStandardScanner [](http://example.com/codeflow?start=20&length=47)

I discovered the `DetectionMetadata.HighEntropy` problem doing other work and this test uncovered the `RotationPeriod` bug. :)

nguerrera (dismissed):

rwoll (approved):

PR #185, opened 2025-04-02
Allow default regex options like NonBacktracking to be disabled

nguerrera:
In the azure-pipelines-agent scenario, we have to support arbitrary .NET regexes, which may be incompatible with non-backtracking. However, there was no way to stop the library from using it when running on .NET 8+. 

This change makes any explicit `RegexOptions` passed to `RegexPattern` constructor or `IRegexEngine`, the definitive full set of options to use. If callers want the default plus some additional options, it is now their responsibility to `|` their additional options onto `RegexDefaults.DefaultOptions`.

`RegexOptions` default arguments are now nullable and null by default. Null indicates "use the default options". Nullable is used to prevent any future change to the defaults from requiring callers to be recompiled. It also deals with a corner case where a portable library could compile against netstandard2.0 and run on net8.0+ and lose `NonBacktracking`.

This involved some refactoring as the regex option defaults were spread in multiple places. They are now consolidated to one place: `RegexDefaults.DefaultOptions` (renamed from `RegexDefaults.DefaultOptionsCaseSensitive`). This consolidation revealed that we were not in fact using `CultureInvariant` even though it was specified there. Now we actually use it by default everywhere.


src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -38,13 +45,14 @@ public RegexPattern(string id,
         DetectionMetadata = patternMetadata;
         RotationPeriod = rotationPeriod;
         Signatures = signatures;
-        m_regexOptions = regexOptions;
+        RegexOptions = (regexOptions | RegexDefaults.DefaultOptions) & ~disabledRegexOptions;
```
nguerrera: This is maybe unnecessarily complicated. I could also just ask the caller to `|` on to `DefaultOptions` if that's what they want. I reasoned this would be a bigger break because previously the options you passed would get augmented by the engine with the DefaultOptions so I was trying to effectively retain that behavior. But I wonder if anyone is actually using this argument now. If I was sure nobody was, I'd go for the simpler approach.

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -18,17 +18,24 @@ public class RegexPattern
 {
     public const string FallbackRedactionToken = "+++";
 
+    /// <summary>Constructs a new instance of the RegexPattern class.</summary>
+    /// <param name="id"> The unique identifier for the pattern.</param>
+    /// <param name="name">The name of the pattern.</param>
+    /// <param name="pattern">The regular expression pattern.</param>
+    /// <param name="patternMetadata">Pattern metadata that describes the pattern.</param>
```
michaelcfanning: >Pattern metadata that describes the pattern [](http://example.com/codeflow?start=38&length=43)

`Metadata that describes the pattern` is sufficient SEV4 #Resolved

nguerrera:
```suggestion
    /// <param name="patternMetadata">Metadata that describes the pattern.</param>
``` #Resolved

File: src/Microsoft.Security.Utilities.Core/RegexPattern.cs (hunk 1 above)
Comment (michaelcfanning) [reply src/Microsoft.Security.Utilities.Core/RegexPattern.cs]: I had a different thought, simply make this a nullable value. If null, prefer the default options, otherwise use the options provided by the caller in toto.

I do *not* believe anyone is using this argument currently, so have at it (in terms of reaching for greatest simplicity). All this stuff about us trying to perform strictly additive stuff or eliding flags is a bit too much supervision. Just trust the caller to make the call and let them specify what they want.

File: docs/ReleaseHistory.md [hunk 3]
```diff
@@ -24,7 +24,13 @@
...
 - BRK: Remove derived `RegexPattern` class properties `ChecksumSeeds`, `EncodeForUrl`, and `KeyLength` as these are not relevant to the literal authoring of equivalent regex patterns in other languages.
 - BRK: Rename `LegacyCommonAnnotatedSecurityKey` to `CommonAnnotatedSecurityKey` and mark this inlined `Microsoft.Security.Utilities` class as internal.
+- BRK: Remove methods on `CachedDotNetRegex` that are not supported by the common `IRegexEngine` interface.
+- BRK: Direct use of `CachedDotNetRegex` no longer forces `RegexOptions.Compiled` and `RegexOptions.NonBacktracking`. It is up to the caller to pass them in if overriding the default argument.
+- BRK: Rename `RegexDefaults.DefaultOptionsCaseSensitive` to `RegexDefaults.DefaultOptions`.
+- BRK: Remove `RegexPattern.DefaultRegexOptions`. Use `RegexDefaults.DefaultOptions` instead.
+- BRK: `RegexOptions.CultureInvariant` is now used by default.
 - NEW: Provide deterministic ordering of properties in GeneratedRegexPatterns/*.json via `DataMember` attributes.
+- NEW: It is now possible to turn off default regex options by passing `disabledRegexOptions` to the `RegexPattern` constructor.
```
michaelcfanning: >regex [](http://example.com/codeflow?start=46&length=5)

Remove this now yes?

src/Tests.Microsoft.Security.Utilities.Core/RegexPatternTests.cs
```diff
@@ -425,4 +425,25 @@ public void RegexPatterns_RaisesArgumentException_OnNullPattern()
             new RegexPattern(Id, Name, DetectionMetadata.HighEntropy, null);
         });
     }
+
+    [TestMethod]
+    public void RegexPattern_RegexOptions_UsesDefault()
+    {
+        var pattern = new RegexPattern("id", "name", DetectionMetadata.None, ".");
+        pattern.RegexOptions.Should().Be(RegexDefaults.DefaultOptions,
+                                         because: "No regex options were passed so default opptions should be used.");
```
michaelcfanning: >opptions [](http://example.com/codeflow?start=91&length=8)

s/be `options` #Resolved

nguerrera:
```suggestion
                                         because: "No regex options were passed so default opptions should be used.");
```
```suggestion
                                         because: "No regex options were passed so default options should be used.");
``` #Resolved

File: src/Tests.Microsoft.Security.Utilities.Core/RegexPatternTests.cs [hunk 5]
```diff
@@ -425,4 +425,25 @@ public void RegexPatterns_RaisesArgumentException_OnNullPattern()
             new RegexPattern(Id, Name, DetectionMetadata.HighEntropy, null);
         });
     }
+
+    [TestMethod]
+    public void RegexPattern_RegexOptions_UsesDefault()
```
michaelcfanning: >RegexPattern_RegexOptions_UsesDefault [](http://example.com/codeflow?start=16&length=37)

Is this test actually passing?

If you pass a non-null default '0' won't that flow through?

nguerrera: It is passing. null is used here, there is no explicit arg. I will add another test for explicit none.

nguerrera: Talked offline, DetectionMetadata.None looked like RegexOptions.None. Improved tests to make it more obvious.

docs/ReleaseHistory.md
nguerrera: yes, thanks

michaelcfanning: we should also add a `BRK` on the conversion to a nullable type for the relevant ctor arg.

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
michaelcfanning: Hey, because you are a scrupulous developer, I suggest you update your PR description to note the interesting subtleties in behavior that drove us landing on the nullable num arg.

michaelcfanning (dismissed):
:shipit:

michaelcfanning (dismissed):

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
nguerrera: Done.

michaelcfanning (approved):
:shipit:

PR #186, opened 2025-04-02
Add dotnet format to `BuildAndTest.ps1`

michaelcfanning:
We now run dotnet format before invoking build in scripts.

As a result, any reformatting changes that occur will break the build (as the source files will be modified).

This change also includes a formatting run.


src/Microsoft.Security.Utilities.Core/AzureSearchIdentifiableKeys.cs
```diff
@@ -1,37 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-using System;
-
-namespace Microsoft.Security.Utilities;
-
-internal sealed class AzureSearchIdentifiableKeys : RegexPattern
```
michaelcfanning, line 8: >AzureSearchIdentifiableKeys [](http://example.com/codeflow?start=22&length=27)

This internal sealed class is entirely unused. It's a vestige of a base class we never used for the admin and query search rules.

src/Microsoft.Security.Utilities.Core/IdentifiableSecrets.cs
```diff
@@ -942,20 +942,20 @@ internal static byte[] GetBase64EncodedSignatureBytes(int keyLengthInBytes,
...
         {
             case 2:
-            {
-                // Clear two bits where the signature will be right-shifted
-                // to align on the base64-encoded 6-bit boundary.
-                mask = 0xfcffffff;
-                break;
-            }
+                {
+                    // Clear two bits where the signature will be right-shifted
```
michaelcfanning, line 946: >Clear [](http://example.com/codeflow?start=23&length=5)

I don't like this indentation scheme. so will try to revert this.

scripts/BuildAndTest.ps1
```diff
@@ -52,10 +52,16 @@ function Exit-WithFailureMessage($scriptName, $message) {
 }
 
 if (-not $NoBuild) {
+    Write-Information "Reformatting code..."
+    dotnet format $RepoRoot\src\Microsoft.Security.Utilities.sln
```
michaelcfanning: >format [](http://example.com/codeflow?start=11&length=6)

Nick, do you know whether a straight `dotnet format` achieves a superset of what happens with `dotnet format style` sub-command (or `dotnet format analyzers`)?

nguerrera:
I find it very surprising when a "build" command overwrites non-generated source files.
 
I would prefer if we had a check like Cask and then you run `dotnet format` to fix. This check shouldn't be part of anything anyone would use as their inner loop, though.

To get an everything gets formatted automatically while I code experience, I strongly prefer format-on-save in IDE. (*) The reason is that when I get into a flow, I do not use the command line BuildAndTest on each iteration. If I commit something that I've only built and tested in the IDE, then run BuildAndTest later then commit something else, I don't want the formatting for the prior commit in the second one.

(*) VS code has the great ability to turn on format-on-save on an individual repo for all users of the repo. Big VS was behind on this last I checked, but there was something coming very soon, so I need to check again. But absent this, I'd prefer manual `dotnet format` outside of inner loop at the very end to get things green in PR validation.

src/.editorconfig
```diff
@@ -212,6 +212,27 @@ dotnet_diagnostic.CA2246.severity=silent
 dotnet_diagnostic.CA2255.severity=silent
 dotnet_diagnostic.CA2234.severity=silent
 dotnet_diagnostic.CA2252.severity=silent
+csharp_using_directive_placement = outside_namespace:silent
```
michaelcfanning, line 215: >csharp_using_directive_placement [](http://example.com/codeflow?start=0&length=32)

This expansion of explicitly specified things resulted simply from me opening .editorconfig in VS to examine certain settings. As a result of these additions, dotnet format subsequently began to enforce the casing indentation I wanted.

So apparently, the literal absence of an explicit entry may be a clue to dotnet format skip certain reformattings.

scripts/BuildAndTest.ps1
nguerrera: I believe it does, yes.

nguerrera: Actually, maybe not. I was thinking I've seen it fix analyzer issues, but they may have been considered "style" analyzers.

src/Microsoft.Security.Utilities.Core/Marvin.NetFramework.cs
```diff
@@ -112,32 +112,32 @@ public static long ComputeHash(byte[] data, ulong seed, int offset, int length)
                 switch (length)
                 {
                     case 0:
-                        p0 += 0x80u;
-                        break;
+                    p0 += 0x80u;
```
nguerrera, line 115: It's now not indenting cases with no braces. I suspect this is not what you want either.

michaelcfanning:
I do not agree with this, it doesn't bother me. Just as I don't mind living with (in fact appreciate) VS automatic reformatting of code when I do things like close a brace. However, I have absolutely no objection to adjusting the dotnet format command to pass `--verify-no-changes` and simply fail the build. Easy, will make you happier and complicate my own life not at all. :)

---
In reply to: [2776080969](https://github.com/microsoft/security-utilities/pull/186#issuecomment-2776080969) [](http://example.com/codeflow?ancestors=2776080969)

src/.editorconfig
nguerrera: I think it would be worth trying the Cask .editorconfig and then tweaking anything you don't like. And then I'd push anything we decide back onto Cask. 

That is what I was intending to do when I filed #153, but I haven't got around to it.

(I'm a broken record, but it's a lot of overhead for stuff like this that we have multiple repos to do it in. I spent a bunch of time on this in Cask and if that wasn't its own repo, it would have provided more value for only marginally more effort to get the settings good enough for all the code.)

michaelcfanning: I was just about to offer your same rant. What I did, in fact, was go look at SPMI and some other projects for a pattern (not CASK). :)

nguerrera:
BuildAndTest failing is not quite what I want either.

We actually agree on IDE reformatting being nice, but I do have a pickier additional opinion where I only want IDE reformatting when I know the repo enforces formatting. Because if it`s not enforced, I do not want random reformatting of existing things to ever happen. I only want to format net new code automagically.

nguerrera:
I managed to get microsoft/typespec to a place where this was never something you had to think about while still giving me the things I want that not everyone else necessarily cares as much as I do about. I was on my way to getting parity with that in Cask (just missing a promised VS feature for parity with VS Code).

scripts/BuildAndTest.ps1
```diff
@@ -53,9 +53,9 @@ function Exit-WithFailureMessage($scriptName, $message) {
 
 if (-not $NoBuild) {
     Write-Information "Reformatting code..."
-    dotnet format $RepoRoot\src\Microsoft.Security.Utilities.sln
+    dotnet format --verify-no-changes $RepoRoot\src\Microsoft.Security.Utilities.sln --verbosity diagnostic
```
nguerrera: Oh wow, it was already doing this? Was something causing it to no-op? I ask because I've never noticed it having anything impact.

scripts/BuildAndTest.ps1
```diff
@@ -53,9 +53,9 @@ function Exit-WithFailureMessage($scriptName, $message) {
 
 if (-not $NoBuild) {
     Write-Information "Reformatting code..."
```
nguerrera: Change to "Checking code formatting..." or something like that.

scripts/BuildAndTest.ps1
nguerrera: Nevermind, was looking at individual commit.

scripts/BuildAndTest.ps1
```diff
@@ -52,10 +52,16 @@ function Exit-WithFailureMessage($scriptName, $message) {
 }
 
 if (-not $NoBuild) {
+    Write-Information "Reformatting code..."
+    dotnet format --verify-no-changes $RepoRoot\src\Microsoft.Security.Utilities.sln --verbosity diagnostic
```
nguerrera: Can we do this in PR validation, but not in BuildAndTest with no opt-in args? In my experience dotnet format is slow.

nguerrera: Let`s leave this for now. We want to stop the formatting bleeding that keeps cropping up on every PR and then when we have that, we can tweak more leisurely. :)

src/.editorconfig
nguerrera: Turns out there's too many things in CASK to absorb quickly, so resolving this. Won't fix on first step forward. :)

src/Microsoft.Security.Utilities.Core/Marvin.NetFramework.cs
nguerrera: We'll fix this separately.

scripts/BuildAndTest.ps1
michaelcfanning: OK. I honor your deep care factor and we will follow your lead unless/until you run up against one of my own bugaboos. Like, C++ style parameter indentation. ;)

nguerrera (dismissed):

scripts/BuildAndTest.ps1
michaelcfanning: Moved this to PR workflow as you suggested.

nguerrera: Even better, thanks!

nguerrera (approved):

PR #190, opened 2025-04-09
Add detection callback to MaskSecrets, make Detection immutable

### LINKED ISSUE #169: Review code for mutable reference types that may get lost in dictionaries

### LINKED ISSUE #172: API to do secret masking and receive detection info in one pass
Currently, you can either use SecretMasker.DetectSecrets or SecretMasker.MaskSecrets. If you use the latter, your input is redacted without returning to you the data about the detections that triggered redactions. 

Consider the scenario where you'd like to perform redaction and log information about the redactions performed. This will be needed in follow-up work after https://github.com/microsoft/azure-pipelines-agent/pull/5154

Possible approaches discussed so far:

1. API that returns both redacted string and detections.
2. API that can use detections to perform redactions.
3. Callback from redaction API with each detection being used for redaction.

nguerrera:
* Add callback to MaskSecrets so detection data can be processed. Fix #172
* Make detection sealed, immutable, and compare equality by reference. Fix #169 
* Optimize away some intermediate allocations in MaskSecrets.

Took on #169 because a mutation in the callback would break things and defending against it would make confusing defensive copies even more confusing.

Then took on the optimization because the algorithm was relying on mutating Detection.Length.


src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -122,7 +122,7 @@ public virtual void AddRegex(RegexPattern regexSecret)
         }
     }
 
-    public string MaskSecrets(string input)
+    public string MaskSecrets(string input, Action<Detection>? onDetection = null)
```
nguerrera: Copilot chose this name, but maybe this isn't so idiomatic for C#. Suggestions? `detectionAction`, perhaps? #Resolved

docs/ReleaseHistory.md
```diff
@@ -12,6 +12,10 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- BRK: `SecretMasker.MaskSecrets` and `ISecretMasker.DetectSecrets` now accept an additional optional parameter. This is binary breaking only for callers, source-breaking as well for implementations of `ISecretMasker`.
+- NEW: `SecreMasker.MaskSecrets` and `ISecretMasker.DetectSecrets` now accept an callback to receive the detection results that are triggering redaction.
+
```
nguerrera, line 20: Struggled here on BRK vs. NEW since it's both so I split them.

michaelcfanning: OK, for BRK we generally want an explicit reference to name + type of added/changed thing. Also, you can use two tags for the same entry, a convention we use elsewhere.

Something like this?

- BRK, NEW: `SecretMasker.MaskSecrets` and `ISecretMasker.DetectSecrets` now accept an `Action<Detection>` callback (with a default value of null) to receive detections that result from the operation. This is binary-breaking only for `SecretMasker` callers and source-breaking as well for `ISecretMasker` implementations.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
michaelcfanning: It's not a bad name but it's true that `action` seems to be a typical term in API, so let's take your suggestion, `detectionAction`

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -142,6 +142,8 @@ public string MaskSecrets(string input)
         Detection? currentDetection = default;
         foreach (Detection detection in detections.OrderBy(x => x.Start))
         {
+            onDetection?.Invoke(detection);
```
michaelcfanning: >onDetection [](http://example.com/codeflow?start=12&length=11)

OK, so we don't assume a need to provide any synchronization here, that would be up to action provider to handle if necessary.

You didn't create all the debt in this file, but I wonder if we should add a doc comment for this method and maybe start addressing some clean-up as we touch various things.

This file is in an unfortunately inconsistent condition but we do have some other doc xml. Maybe just add a good summary for this changed method?

michaelcfanning: OK, so you must have thought this through and believe that firing a callback for any discrete detection (even in overlapping cases) is a good thing.

Rather than having the callbacks align precisely with masked regions. 

I tend to think that's right. It does raise the detailed point that we don't do anything explicitly related to if/when we handle the contiguous or overlapping cases. The callback receiver could do that additional work if they wanted. Just thinking aloud.

 #Resolved

docs/ReleaseHistory.md
nguerrera: Ah, perfect. Thanks!

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Sure, I'll add a doc.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -755,6 +755,35 @@ public void SecretMasker_DistinguishLiteralAndRegexRedactionTokens()
         Assert.AreEqual("zzzxyyyab12", result);
     }
 
+    [TestMethod]
+    public void SecretMasker_ProvidesDetectionsViaCallback()
```
michaelcfanning, line 760: >SecretMasker_ProvidesDetectionsViaCallback [](http://example.com/codeflow?start=16&length=42)

Related to other discussion. You should consider updating this test (or adding a new one) that provokes both contiguous and overlapping matches (because in our implementation we still require the code to fire a callback for every individual match.

The regexes below would do well for this

'patter'
'pattern'
'n1'

with data like 

'yada patter pattern pattern1'

The masker should collapse this to
'yada +++ +++ +++'

But we should get six callbacks?

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: On synchronization, yes, but now I'm again annoyed that Detection is mutable. Maybe it's time to bite that bullet. If detection were mutated here, it could break the algorithm.

nguerrera: Yes, I did think about it, and that was my conclusion too. My immediate use case of logging what was found that triggered masking does not care about whether they were contiguous, disjoint, or overlapping.

nguerrera: OK, I bit the bullet, see updated title description and code changes.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
nguerrera: Will do

nguerrera: Hmm, there's an annoying edge case here.

If detections overlap and have the same start index, then which redaction token is supposed to win? I think before my change, it was whichever detection came first, which is actually not well-defined either because it relies on hashset enumeration order. With my change, I added an unstable sort to the mix. Blah.

Maybe we should be sorting by redaction token too? I'll do that.

docs/ReleaseHistory.md
```diff
@@ -12,6 +12,12 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- BRK, NEW: SecretMasker.MaskSecrets and ISecretMasker.DetectSecrets now accept an Action<Detection> callback (with a default value of null) to receive detections that result from the operation. This is binary-breaking only for SecretMasker callers and source-breaking as well for ISecretMasker implementations.
```
michaelcfanning: >SecretMasker [](http://example.com/codeflow?start=12&length=12)

all the enquoting chars for named things are missing from this entry (though they are present for others).

SEV3 `SecretMasker.MaskSecrets`, etc.

docs/ReleaseHistory.md
```diff
@@ -12,6 +12,12 @@
...
+- BRK: `Detection` is now sealed, and immutable. All property setters, and the copy constructor is removed.
+- BRK: `Detection` is no longer overrides Equals and GetHashCode nor implements `IEquatable<Detection>`. Equality comparison of `Detection` will now be based on reference equality.
```
michaelcfanning: >Equals [](http://example.com/codeflow?start=42&length=6)

`Equals` and `GetHashCode`. this standard isn't a *foolish* consistency and so it isn't a hobgoblin of a little mind. nay, it is a prudent consistency. SEV3

docs/ReleaseHistory.md
nguerrera: I think GitHub helpfully dropped them when I cut and paste your text. Thanks, GitHub. :) Will fix.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
nguerrera: Outstanding feedback here because I broke this when I optimized how we process the list of detections. Fixed. 😊

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -122,62 +122,80 @@ public virtual void AddRegex(RegexPattern regexSecret)
...
+    /// <summary>
+    /// Masks secrets detected by <see cref="DetectSecrets(string)"/> in the
+    /// input, replacing them with <see cref="Detection.RedactionToken"/>.
+    /// </summary>
+    /// <remarks>
+    /// When secrets overlap or are adjacent to each other, the entire range
+    /// encompassing them is masked and the leftmost detection's redaction token
+    /// is used. If there is more than one overlapping leftmost detection, then
+    /// the redaction token among them that sorts first by ordinal
+    /// case-sesstring comparison is used.
```
michaelcfanning: >case-sesstring [](http://example.com/codeflow?start=8&length=14)

? 'ordinal case-sensitive string comparison'?

michaelcfanning (approved):
:shipit:

docs/ReleaseHistory.md
```diff
@@ -12,6 +12,12 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- BRK, NEW: `SecretMasker.MaskSecrets` and `ISecretMasker.DetectSecrets` now accept an `Action<Detection>` callback (with a default value of `null`) to receive detections that result from the operation. This is binary-breaking only for `SecretMasker` callers and source-breaking as well for `ISecretMasker` implementations.
+- BRK: `Detection` is now sealed, and immutable. All property setters and the copy constructor are removed.
```
michaelcfanning: >, [](http://example.com/codeflow?start=32&length=1)

SEV5 you can remove this comma. Gosh! I see there is a SEV5 after all!

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Yikes. Fixed.

michaelcfanning (changes requested):
:clock1:

michaelcfanning (approved):
:shipit:

PR #191, opened 2025-04-22
Fix build warning, enable warn-as-error

nguerrera:


src/Directory.WarningsAsErrors.props
```diff
@@ -0,0 +1,28 @@
+<Project>
+  <!--
+    NOTE: We only turn on warnings-as-errors in Release because it hurts
```
michaelcfanning, line 3: >Release [](http://example.com/codeflow?start=48&length=7)

love it nice work, right balance

src/Directory.WarningsAsErrors.props
```diff
@@ -0,0 +1,28 @@
...
+    <!-- Treat compiler warnings as errors. -->
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+
+    <!-- Treat all build warnings as errors -->
+    <MSBuildTreatWarningsAsErrors>$(TreatWarningsAsErrors)</MSBuildTreatWarningsAsErrors> 
+
+    <!--
+      https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1901-nu1904
+
+      NU190X: Package 'A.B.C' N.N.N has a known (low|moderate|high|critical)
```
michaelcfanning, line 17: >Package [](http://example.com/codeflow?start=14&length=7)

OK, fine, yes. We could go this route, I'm entirely comfortable with it. I will note that NU1904 is a code for critical severity and there might perhaps be a case for permitting this highest level of severity to continue to break us.

If we did this, maybe we could just set a NuGetAuditLevel to critical as an approach. The downside here may be that the current crop of warnings may be less visible.

To beat this whole thing to death, I'll also note that we could eliminate the `WarningsNotAsErrors` thing and create a new props file to hold explicit suppressions of various advisories, via `NuGetAuditSuppress`. This would factor our technical debt in the area explicitly in a file, helpful when making a push to close on it. A downside here is that working against this baseline, we'd need to accept the fact that alerts could materialize at any time (i.e., accept the costs of breaking unexpectedly and requiring a dev to add a suppression in the baseline if need be).

This seems like as good a time as any to ask the general question, what's our attitude around allowing technical debt to accumulate? We are in the worst possible situation for this kind of problem, in my opinion. As a library, we want to put as few requirements/limits as is reasonable for upstream callers. OTOH, we want to be as secure as possible.

michaelcfanning (approved):
:shipit:

src/Directory.WarningsAsErrors.props
nguerrera: Interesting. First thing to note I've spread this exact file into many places over the last few months. Yay, polyrepos, lol.  If we decide to do it differently, we'll have to scrub! I'll think about it, but my (not so strong) opinion is that it's still better to retain the property that builds don't spontaneously break themselves. There should be other alerts for vulnerable packages.

PR #192, opened 2025-04-22
Merge IdentifiableScan into SecretMasker

### LINKED ISSUE #142: Reconsider pervasive use of iterators and `IEnumerable<T>` returns
`IEnumerable<T>` is the most convenient collection parameter type for callers because it lets them pass any collection. However, it is the *least* convenient collection *return* type because it gives you back something that has only the lowest common denominator of all collection API.

For properties which return constant collections, like `WellKnownRegexPatterns.*`, it feels especially out-of-place to me. `IReadOnlyList<T>` or even `ImmutableArray<T>` would be better. I also don't think using `yield` to concatenate two such lists into one is any better than a one-time cost to realize a concatenated collection. These lists aren't huge.

For API like `DetectSecrets`, I can see more utility as a caller gains the ability to interrupt the operation.  However, much like `async`, I think it's helpful if API that are truly backed by enumerable-only collections (such as with iterators) indicate this in their name. E.g. `EnumerateSecretDetections`. This prefix has a lot of prior art in the BCL. The reason for this is that callers should be more careful with these than plain collections. All too often, I see redundant enumeration. For example, with a `.Count()` before the enumerating again in a loop.

Aside from being more cumbersome for callers, iterators are cumbersome at development time too. It's annoying when the debugger doesn't step straight into where you expect and when you have to force it to enumerate and show you all the items.

### LINKED ISSUE #154: Merge IdentifiableScan into SecretMasker
There's no reason to have competing entry points where one of them is faster for certain patterns. We should have one entry point (SecretMasker) that does the fastest thing it can for the patterns it's given.

### LINKED ISSUE #171: IdentifiableScan does not dispose backup secret masker
IdentifiableScan is IDisposable, and owns a SecretMasker for patterns that aren't supported by HighPerformanceScanner, but it does not dispose the SecretMasker when it is disposed.

NOTE: This predates #132. Dispose is now empty, but prior to that, it only cleaned up the rust resources and did not dispose the backup secret masker.

nguerrera:
Merge IdentifiableScan into SecretMasker

- BRK, PRF: `IdentifiableScan` has been removed and its functionality has been merged into `SecretMasker` which now scans the patterns it's given as fast as possible, which will be significantly faster for identifiable key patterns.
- BRK: `SecretMasker.Clone` has been removed.
- BRK: `WellKnownRegexPatterns.*Iterator` methods have been removed. Use the corresponding `WellKnownPatterns.*` properties instead.
- BRK: `WellKnownRegexPatterns.*` properties are now of stronger type`IReadOnlyList<RegexPattern>` instead of `IEnumerable<RegexPattern>`.
- BRK: `SecretMasker.SyncObject` is now a read-only property instead of a public mutable field.
nguerrera marked this conversation as resolved.
- NEW: `SecretMasker` is now capable of finding identifiable keys without relying on delimiting characters`.
- PRF: Add high-performance scanning for additional patterns: `SEC101/031.NuGetApiKey`, `SEC101/050.NpmAuthorKey`, `SEC101/110.AzureDatabricksPat`, and `SEC101/565.SecretScanningSampleToken`.
- FPS: Prevent `SEC101/105.AzureMessagingLegacyCredentials` from firing on non-legacy keys.

Also:
- Keep a test hook to use the non-high-performance path in tests and benchmarks.
- Improve benchmark with parameterization and an additional non-prefixed case to validate that the high-performance scanner does not regress perf for small input. In fact, it is still noticeably faster on small inputs, but less dramatically than on large inputs.
- Remove dead code in AzureStorageAccountLegacyCredentials.

Fix #154
Fix #171
Fix #142


src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -20,17 +20,13 @@ namespace Microsoft.Security.Utilities;
...
+    private Dictionary<string, IList<RegexPattern>> _highPerformanceSignatureToPatternsMap;
 
-    public static Version Version => RetrieveVersion();
-
-    internal static Version RetrieveVersion()
-    {
-        var version = new Version(ThisAssembly.AssemblyFileVersion);
-        return new Version(version.Major, version.Minor, version.Build);
-    }
+    public static Version Version { get; } = new Version(ThisAssembly.AssemblyFileVersion);
```
nguerrera: release note #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -402,28 +490,47 @@ public virtual SecretMasker Clone()
...
 
-
-    private readonly bool m_generateCorrelatingIds;
+    private readonly bool _generateCorrelatingIds;
     public HashSet<LiteralEncoder> LiteralEncoders { get; }
     public HashSet<SecretLiteral> EncodedSecretLiterals { get; }
     public HashSet<SecretLiteral> ExplicitlyAddedSecretLiterals { get; }
 
-    public ReaderWriterLockSlim SyncObject = new(LockRecursionPolicy.NoRecursion);
+    public ReaderWriterLockSlim SyncObject { get; } = new(LockRecursionPolicy.NoRecursion);
```
nguerrera, line 501: release note

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -348,23 +393,35 @@ public IEnumerable<Detection> DetectSecrets(string input)
             yield break;
         }
 
-        if (RegexPatterns.Count == 0 &&
-            EncodedSecretLiterals.Count == 0 &&
-            ExplicitlyAddedSecretLiterals.Count == 0)
-        {
-            yield break;
-        }
-
```
nguerrera, line 357: This should have been in the read lock, but this condition to check if we have a degenerate secret masker with nothing added to it doesn't seem worth optimizing for.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -71,6 +69,12 @@ protected SecretMasker(SecretMasker copy)
         try
         {
             copy.SyncObject.EnterReadLock();
+
+            _regexEngine = copy._regexEngine;
+            _highPerformanceScanner = copy._highPerformanceScanner;
+            _highPerformanceSignatureToPatternsMap = copy._highPerformanceSignatureToPatternsMap;
+            _generateCorrelatingIds = copy._generateCorrelatingIds;
```
nguerrera: regexEngine and generate correlating IDs were missing, add test coverage. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: actually, no, this is non-breaking.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Still working on tests here, just fixed more bugs in it.

I wonder if we really need Clone. We inherited it from the VSO interface, but I'm not sure that pipeline agent actually uses it...

docs/ReleaseHistory.md
```diff
@@ -16,7 +16,14 @@
 - BRK, NEW: `SecretMasker.MaskSecrets` and `ISecretMasker.DetectSecrets` now accept an `Action<Detection>` callback (with a default value of `null`) to receive detections that result from the operation. This is binary-breaking only for `SecretMasker` callers and source-breaking as well for `ISecretMasker` implementations.
 - BRK: `Detection` is now sealed and immutable. All property setters and the copy constructor are removed.
 - BRK: `Detection`  no longer overrides `Equals` and `GetHashCode` nor implements `IEquatable<Detection>`. `Detection.Equals` and `Detection.GetHashCode` will therefore now be based on reference equality.
+- BRK, PRF: `IdentifiableScan` has been removed and its functionality has been merged into to `SecretMasker` which now scans the patterns its given as fast as possible, which will be significantly faster for identifiable key patterns.
```
michaelcfanning: >its [](http://example.com/codeflow?start=138&length=3)

`it's`. 
signed, the `its vs. it's` police #Resolved

docs/ReleaseHistory.md
```diff
@@ -16,7 +16,14 @@
...
+- BRK: `WellKnownRegexPatterns.*Iterator` methods have been removed. Use the corresponding `WellKnownPatterns.*` properties instead.
+- BRK: `WellKnownRegexPatterns.*` properties are now of stronger type`IReadOnlyList<RegexPattern>` instead of `IEnumerable<RegexPattern>`.
+- BRK: `SecretMasker.SyncObject` is now a read-only property instead of a public mutable field.
```
michaelcfanning, line 23: >SecretMasker.SyncObject [](http://example.com/codeflow?start=8&length=23)

holy crud. #Resolved

docs/ReleaseHistory.md
```diff
@@ -16,7 +16,14 @@
...
+- PRF: Add high-performance scanning for additional patterns: `SEC101/031.NuGetApiKey`, `SEC101/050.NpmAuthorKey`, `SEC101/110.AzureDatabricksPat`, and `SEC101/565.SecretScanningSampleToken`.
 - PRF: Fewer intermediate allocations are performed by `SecretMasker.MaskSecrets`.
+- NEW: `SecretMasker` is now capable of finding identifiable keys without relying on delimiting characters`.
```
michaelcfanning: >NEW [](http://example.com/codeflow?start=2&length=3)

a `NEW` entry should precede all `PRF`. Also there is a trailing backtick at the end of this release note. #Closed

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -5,42 +5,85 @@
...
+        // base64 alphabet char is equally likely to appear in the input, making
+        // signature sniffing harder.
+        HardPrefix,
+
+        // This is close to the best case scenario. A char that does not appear
+        // in any identifiable key signature is repeated in the prefix, making
+        // signature sniffing easier.
+        EasyPrefix,
+
+        // This will scan secrets without prefixing them. Testing the performance against
```
michaelcfanning: >Testing [](http://example.com/codeflow?start=58&length=7)

Not a complete sentence. 'This is useful for testing performance against small inputs.' ?? #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -5,42 +5,85 @@
...
+        // small inputs.
+        NoPrefix,
+    }
 
-        protected List<string> RandomExamples { get; }
-        protected List<string> NonRandomExamples { get; }
+    public abstract class SecretMaskerDetectionBenchmarks
+    {
+        // The size of randomized data to add as a prefix for every secret for
+        // the EasyPrefix and HardPrefix cases. This is intended to make
```
michaelcfanning: >EasyPrefix [](http://example.com/codeflow?start=15&length=10)

You could cref this and  `HardPrefix, why not. #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -5,42 +5,85 @@
...
+        // positive hits less concentrated in the profiling.
+        public const int SecretPrefixSize = 10 * 1024;
```
michaelcfanning: >1024 [](http://example.com/codeflow?start=49&length=4)

ok, so default filler material is 10K? that still seems awfully concentrated compared to a real world scenario. Also this assumes this number is `InBytes` (you might consider adding an explanatory size suffix to the name, I always find this useful). #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -5,42 +5,85 @@
...
-        // for every secret. This is intended to make positive
-        // hit less concentrated in the profiling.
-        private const int SecretPrefixSize = 100 * 1024;
+        DotNet,
+        RE2,
+    }
+
+    public enum Case
+    {
+        // This is close to the worst case scenario for large input. Every
```
michaelcfanning: >is [](http://example.com/codeflow?start=16&length=2)

maybe add these as simple `<summary>` elements? SEV4, why not #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -5,42 +5,85 @@
...
         // Use a fixed seed to ensure different runs use the same data.
-        private readonly Random rng = new Random(Seed: 42);
+        private readonly Random _rng = new Random(Seed: 42);
 
-        protected SecretMaskerDetectionBenchmarks()
-        {
-            RandomExamples = GenerateExamples(random: true);
-            NonRandomExamples = GenerateExamples(random: false);
-        }
+        protected List<string> Examples { get; private set; } = null!;
```
michaelcfanning, line 84: >List [](http://example.com/codeflow?start=18&length=4)

SEV4 perhaps `IList`? is it possible someone might want to provide a sorted list for example? #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: It's InChars, technically. 

It was 100KB, but it's painful to iterate on the benchmark when testing that much.

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -5,42 +5,85 @@
...
                 var data = new byte[size / 4 * 3]; // Account for base64 encoding overhead.
-                rng.NextBytes(data);
+                _rng.NextBytes(data);
                 prefix = Convert.ToBase64String(data);
             }
             else
             {
-                // This is close to the best case scenario. This char does not appear in
-                // any identifiable key signature, making signature sniffing easier.
                 prefix = new string('%', size);
```
michaelcfanning: >'%' [](http://example.com/codeflow?start=36&length=3)

SEV4 maybe change this to a unicode character or something more guaranteed never to appear in a regex, for example `\t`. someday, perhaps a range of chars, which includes things like `\0` could be interesting (but then again maybe this data diversity isn't helpful in the foreseeable future) #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -52,10 +95,10 @@ private string GeneratePrefix(int size, bool random)
             return prefix;
         }
 
-        private List<string> GenerateExamples(bool random)
+        private List<string> GenerateExamples(bool random, int prefixSize)
```
michaelcfanning: `prefixSizeInBytes` or what? #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: ~Can you rephrase this suggestion? I'm not sure what you mean.~

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: Yeah, these should all be doc comments. I copied an existing comment that wasn't and spread it around some more. Will fix.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_106_AzureStorageAccountLegacyCredentials.cs
```diff
@@ -27,103 +25,5 @@ public override Tuple<string, string> GetMatchIdAndName(string match)
...
             return base.GetMatchIdAndName(match);
         }
-
-        /// <summary>
-        /// Given a string, validate that it has a valid checksum.
-        /// </summary>
-        /// <param name="input">String to checksum validate.</param>
-        /// <param name="magicNumber">Magic number to use to validating checksum.</param>
-        /// <returns>True if checksum is valid.</returns>
-        private static bool IsChecksumValid(string input, uint magicNumber)
```
michaelcfanning, line 37: >IsChecksumValid [](http://example.com/codeflow?start=28&length=15)

wha? stale ADO PAT code stuck in here? crazy. #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_565_SecretScanningSampleToken.cs
```diff
@@ -23,6 +18,15 @@ public SecretScanningSampleToken()
             Signatures = new HashSet<string>(["ab85"]);
         }
 
+#if HIGH_PERFORMANCE_CODEGEN
+        IEnumerable<HighPerformancePattern> IHighPerformanceScannableKey.HighPerformancePatterns => [
+            new(signature: "ab85",
+                scopedRegex: "^secret_scanning_.{4}fc6f8d7638cf1c11da812da308d43_[0-9A-Za-z]{5}",
```
michaelcfanning, line 24: >.{4} [](http://example.com/codeflow?start=47&length=4)

why are you permitting any characters here? #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_565_SecretScanningSampleToken.cs
```diff
@@ -5,13 +5,8 @@
 
 namespace Microsoft.Security.Utilities
 {
-    public class SecretScanningSampleToken : RegexPattern
+    public class SecretScanningSampleToken : RegexPattern, IHighPerformanceScannableKey
     {
-        /// <summary>
-        /// Detect 32-character Azure Active Directory client application legacy credentials.
```
michaelcfanning, line 11: >Azure [](http://example.com/codeflow?start=32&length=5)

wow, you sought and destroyed lots of copy and paste nonsense, thanks. #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_565_SecretScanningSampleToken.cs
nguerrera: Because the scoped regex only runs if the signature has matched already. It's a bit of a micro-optimization here, but in other cases this lets us share regexes that would differ only by signature.

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: It's in chars. Maybe I should replace size with length throughout?

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_565_SecretScanningSampleToken.cs
nguerrera: The rust code also does this in its hand-written routines that are the moral equivalent of these "scoped" regexes.

src/Microsoft.Security.Utilities.Core/CompiledHighPerformancePattern.Generated.cs
```diff
@@ -32,6 +32,11 @@ partial class CompiledHighPerformancePattern
     /*lang=regex*/ private const string RawRegex5 = """^[a-zA-Z0-9+/]{33}.{4}[A-P][a-zA-Z0-9+/]{5}=""";
     /*lang=regex*/ private const string RawRegex6 = """^[a-zA-Z0-9+/]{76}.{4}[a-zA-Z0-9+/]{5}[AQgw]==""";
     /*lang=regex*/ private const string RawRegex7 = """^[a-zA-Z0-9+/]{42}.{4}[A-D][a-zA-Z0-9+/]{5}""";
+    /*lang=regex*/ private const string RawRegex8 = """^.{3}[a-z2-7]{43}""";
+    /*lang=regex*/ private const string RawRegex9 = """^.{3}[A-Z2-7]{43}""";
+    /*lang=regex*/ private const string RawRegex10 = """^.{4}[0-9a-f\-]{32,34}""";
```
michaelcfanning, line 37: >{4}[0-9a-f\-]{32,34} [](http://example.com/codeflow?start=58&length=20)

i'm missing the context for adding this particular pattern, why do we need it? will keep looking to see if I can answer my own question.  my eye caught this pattern initially because there's no similar pattern for uppercase alpha. on surface, this sorta looks like it could be a hex 32-byte value with or without hyphens but if so, the length checks don't make sense (there are 4 hyphens to elide in a GUID). #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
```diff
@@ -36,6 +36,22 @@ public override Tuple<string, string> GetMatchIdAndName(string match)
             return base.GetMatchIdAndName(match);
         }
 
+#if HIGH_PERFORMANCE_CODEGEN
+        IEnumerable<HighPerformancePattern> IHighPerformanceScannableKey.HighPerformancePatterns => [
+            new(signature: "oy2",
+                scopedRegex: @"^.{3}[a-z2-7]{43}",
```
michaelcfanning, line 42: >.{3} [](http://example.com/codeflow?start=32&length=4)

same question, will stop asking it, why do we not recap the actual signature here? #Resolved

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -20,7 +20,7 @@ public class SecretMaskerTests
     public void SecretMasker_PreciselyClassifiedSecurityKeys_Detections()
     {
         ValidateSecurityModelsDetections(WellKnownRegexPatterns.PreciselyClassifiedSecurityKeys,
-                                         preciseClassifications: false);
```
michaelcfanning, line 23: >false [](http://example.com/codeflow?start=65&length=5)

wah? #Resolved

michaelcfanning (dismissed):
:shipit:

src/Microsoft.Security.Utilities.Core/CompiledHighPerformancePattern.Generated.cs
nguerrera: This is the generated code for the `AzureDatabricksPat` scoped regex.

I believe it follows from the pre-existing general regex for that correctly. Is there something wrong with that or am I missing a divergence from it?

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
nguerrera: resolving this one since I answered elsewhere.

src/Microsoft.Security.Utilities.Core/CompiledHighPerformancePattern.Generated.cs
nguerrera: Hmm, looking at SPMI and the test cases here, it seems it should be exactly 32 `[0-9a-f]` followed by optional `-[0-9]` ?

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: I'd prefer to leave this. The only use case is in a derived benchmark that wants to bench scanning the examples with a different scanner. (Which is done in the internal benchmark). In that case, I want enumeration of the examples to take as little of the benchmark time as possible.

nguerrera:
(Re-requested review since it was indicated offline that some things still needed to be checked.)

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
```diff
@@ -28,25 +29,48 @@ namespace Microsoft.Security.Utilities;
...
 {
+    private Dictionary<int, CompiledHighPerformancePattern> _patternsByPackedSignature = new();
+
 #if NET9_0_OR_GREATER
-    private SearchValues<string> _signatures;
+    private static SearchValues<string> s_emptySignatures = SearchValues.Create(Array.Empty<string>(), StringComparison.Ordinal);
+    private SearchValues<string> _signatures = s_emptySignatures;
 #elif NET8_0_OR_GREATER
-    private SearchValues<char> _signatureStarts;
+    private static SearchValues<char> s_emptySignatureStarts = SearchValues.Create(Array.Empty<char>());
```
michaelcfanning: >Create [](http://example.com/codeflow?start=76&length=6)

SEV3: my assumption here is that this code path does not support adding the `StringComparison.Ordinal` argument? #Resolved

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
```diff
@@ -28,25 +29,48 @@ namespace Microsoft.Security.Utilities;
...
-    private char[] _signatureStarts;
+    private char[] _signatureStarts = Array.Empty<char>();
 #endif
-    private Dictionary<int, CompiledHighPerformancePattern> _patternsByPackedSignature;
+
+    public HighPerformanceScanner()
+    {
+    }
 
     public HighPerformanceScanner(IEnumerable<CompiledHighPerformancePattern> patterns)
```
michaelcfanning: >HighPerformanceScanner [](http://example.com/codeflow?start=11&length=22)

Here's something small to discuss, we could simply make this ctor have an optionally available parameter. on doing this, though, the class loses its parameterless constructor, which has impact for generics, at least.

Any thoughts on this? SEV5 just discussing in general. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
michaelcfanning: How can we build confidence to zap this? Would be good. I don't see any MS docs for this type. I very much suspect we won't have serious compatibility issues simplifying it. We could try to go through some process of marking it obsolete, etc. Better, perhaps we could agree to a scream test, first simplify accepting that we might need to quickly stub in something to restore if need be. Probably a good bet that we won't need to.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_565_SecretScanningSampleToken.cs
michaelcfanning: OK, so in general, you are saying this is a good practice that futures proofs allowing us to share regexes.

A different lens would say do *not* add this loosening unless/until there is regex sharing. certainly there shouldn't be for this rule (and several others with the pattern). 

but the other (small) part of your case is that generalizing this piece of the regex allows a microoptimization in the regex itself? i.e., simplifying it/making it faster (in some non-zero but potentially small way)?

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -348,23 +362,38 @@ public IEnumerable<Detection> DetectSecrets(string input)
...
+            {
+                foreach (HighPerformanceDetection highPerformanceDetection in _highPerformanceScanner.Scan(input))
+                {
+                    string found = input.Substring(highPerformanceDetection.Start, highPerformanceDetection.Length);
+                    foreach (RegexPattern pattern in _highPerformanceSignatureToPatternsMap![highPerformanceDetection.Signature])
+                    {
+                        Detection? detection = FinalizeHighPerformanceDetection(highPerformanceDetection, found, pattern);
+                        if (detection != null)
+                        {
+                            yield return detection;
```
michaelcfanning, line 336: >return [](http://example.com/codeflow?start=34&length=6)

ok, again just chatting here. the reality is that this code path could in theory permit a 'yield break' after returning the detection, because we should never have results shared between patterns of this kind.

i'm tempted to propose it.

i'm also tempted for us to consider some sorting mechanism for checks that allows us to reliably short-circuit anytime we detect the first match.

the benefit here would be that we would prevent double-firing for specific vs. general patterns (e.g., legacy Azure Storage key vs. a rule that simply detect base64).

the downsides may be that this approach might arguably mask some subtle bugs. also, is it a bug or a feature that if you enable only the general rules that they would fire on something that's more specific (but not when that more specific rule was enabled)?

in general, I think that would be ok. basically, whatever the rule set, allow one thing to fire, the first thing. thoughts? #Pending

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -17,95 +19,66 @@ public static class WellKnownRegexPatterns
...
-    };
+    public static IReadOnlyList<RegexPattern> UnclassifiedPotentialSecurityKeys { get; } =
+        CreateList(new UnclassifiedJwt(),
+                   new UrlCredentials(),
+                   new LooseSasSecret(),
+                   new OAuth2BearerToken(),
+                   new Unclassified32ByteBase64String(),
+                   new Unclassified64ByteBase64String(),
+                   new AadClientAppLegacyCredentials34(),      // SEC101/101 legacy generated passwords.
+                   new Pkcs12CertificatePrivateKeyBundle(),
```
michaelcfanning, line 30: >Pkcs12CertificatePrivateKeyBundle [](http://example.com/codeflow?start=23&length=33)

strictly speaking, this is a classified secret. the issue here is one of confidence level. btw - looking at the regex, we could certainly improve accuracy.

this rule is also interesting to discuss because we could definitely drive our confidence rate to a very high degree if we're willing to attempt deserializing this data. i'm assuming this should be possible without pulling in unwanted deps like bouncycastle, etc., (i.e., with built-in .net crypto support).

I opened a bug on this. The question to you is whether you want this precisely classified (but medium confidence thing) in this collection.
 #Resolved

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -17,95 +19,66 @@ public static class WellKnownRegexPatterns
...
-    public static IEnumerable<RegexPattern> HighConfidenceThirdPartySecurityModels { get; } = new List<RegexPattern>
-    {
-        new NpmAuthorKey(),
-        new SecretScanningSampleToken(),
-    };
+    public static IReadOnlyList<RegexPattern> UnclassifiedPotentialSecurityKeys { get; } =
+        CreateList(new UnclassifiedJwt(),
+                   new UrlCredentials(),
+                   new LooseSasSecret(),
+                   new OAuth2BearerToken(),
```
michaelcfanning, line 26: >OAuth2BearerToken [](http://example.com/codeflow?start=23&length=17)

We should add `Unclassified` as a name for all unclassified things, e.g., `UnclassifiedOAuth2BearerToken`. and `UnclassifiedSasHmac` (my rewrite of `LooseSasSecret`)

[https://github.com/microsoft/security-utilities/issues/195](https://github.com/microsoft/security-utilities/issues/195) #Resolved

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -17,95 +19,66 @@ public static class WellKnownRegexPatterns
...
+                   new Unclassified32ByteBase64String(),
+                   new Unclassified64ByteBase64String(),
+                   new AadClientAppLegacyCredentials34(),      // SEC101/101 legacy generated passwords.
```
michaelcfanning: This rule really is questionable. It flags any pattern of requisite length that is alphanumeric and which contains at least *one* of the unreserved URL symbols, hyphen, underscore, dot and tilde.

It is, according to the rule name, a pattern that proposes an actual classification, i.e., it professes to be an AAD client app legacy password. And so again isn't strictly unclassified (the issue being that it tends to raise a problematic set of false positives). #Resolved

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -17,95 +19,66 @@ public static class WellKnownRegexPatterns
...
+                   new AzureBatchIdentifiableKey(),
+                   new AzureMLWebServiceClassicIdentifiableKey(),
+                   new AzureApimIdentifiableDirectManagementKey(),
+                   new AzureApimIdentifiableSubscriptionKey(),
+                   new AzureApimIdentifiableGatewayKey(),
+                   new AzureApimIdentifiableRepositoryKey(),
+                   new AzureCacheForRedisIdentifiableKey(),
+                   new AzureContainerRegistryIdentifiableKey(),
+                   new NuGetApiKey(),
+                   new AdoLegacyPat(),                         // SEC101/102
```
michaelcfanning: >SEC101/102 [](http://example.com/codeflow?start=66&length=10)

i'm not happy with our partial rendering of the opaque ids, but you didn't create this issue you just moved it about. :) #Resolved

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -17,95 +19,66 @@ public static class WellKnownRegexPatterns
...
+
+    public static IReadOnlyList<RegexPattern> HighConfidenceThirdPartySecurityModels { get; } =
+        CreateList(new NpmAuthorKey(),
+                   new SecretScanningSampleToken());
+
+    public static IReadOnlyList<RegexPattern> PreciselyClassifiedSecurityKeys { get; } =
+        CombineLists(HighConfidenceMicrosoftSecurityModels,
+                     HighConfidenceThirdPartySecurityModels);
+
+    public static IReadOnlyList<RegexPattern> SecretStoreClassificationDetections { get; } =
```
michaelcfanning, line 78: >SecretStoreClassificationDetections [](http://example.com/codeflow?start=46&length=35)

this is intended capture the rules we would want to run when enumerating a secret store, but the idea is unrealized and unconsumed, lets zap it for now. #Resolved

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
```diff
@@ -17,95 +19,66 @@ public static class WellKnownRegexPatterns
...
+                   new AzureEventGridIdentifiableKey());
+
+    public static IReadOnlyList<RegexPattern> DataClassification { get; } =
+        CreateList(new IPv4(),
+                   new IPv6(),
+                   new Float(),
+                   new Integer(),
+                   new GuidValue());
+
+    public static IReadOnlyList<RegexPattern> HighConfidenceThirdPartySecurityModels { get; } =
```
michaelcfanning, line 70: >HighConfidenceThirdPartySecurityModels [](http://example.com/codeflow?start=46&length=38)

I think we should zap this `ThirdParty` business, it's not helpful. Strictly speaking, MS doesn't own the databricks model either, so it's odd to have that one over in the other bucket. Let's just simplify to the high confidence bucket. We can add metadata later on the allocating company if it's helpful. #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
```diff
@@ -36,6 +36,22 @@ public override Tuple<string, string> GetMatchIdAndName(string match)
             return base.GetMatchIdAndName(match);
         }
 
+#if HIGH_PERFORMANCE_CODEGEN
```
michaelcfanning, line 39: >HIGH_PERFORMANCE_CODEGEN [](http://example.com/codeflow?start=4&length=24)

OK, so here's the deal on this one, these secrets actually accept mixed case. NuGet somewhere normalizes keys on receipt before attempting validation. I just confirmed this with a test key, you can change absolutely any key case and things continue to work.

that means you can converge on a single pattern which simply accepts `[a-z2-7]`

alternately you can use my radically over-engineered regex here it is:

> ^.{3}[a-pA-P][0-9a-zA-Z]{15}[aqAQ][0-9a-zA-Z]{11}[euEU][bdfhjlnprtvxz357BDFHJLNPRTVXZ][a-pA-P][0-9A-Za-z]{11}[aeimquy4AEIMQUY]

the problem of accepting my regex, of course, is that the test examples would need to be updated to conform more closely to the mechanism nuget uses to build its keys. there is an embedded guid in the encoded key, from which this over-engineered thing derives (because a guid has reserved bits that limits their expression in the nuget base32 encoded form). #Resolved

michaelcfanning: the other issue with accepting my regex is simply that you hate it.

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: Ah, `<summary>` was not in backticks and treated as html in your comment. Edited to render correctly, and now I get it, and yes, will do.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
nguerrera: I'm going to leave behavior changes to the rules out of scope for this large PR and I will file an issue.

Note that the behavior is unchanged from before since there is deliberate code to reject mixed case right above this. Comment claims "A match with a mix of case is more likely to be base64 than a real API key."

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_565_SecretScanningSampleToken.cs
nguerrera: Yes, to the last (small) part, you can actually see the code getting improved if you go-to-definition on the methods with GeneratedRegexAttribute.

My inclination here was to keep all of the high-performance patterns consistent in skipping the signature rather than make individual judgment calls. I don't feel super strongly about it, but that still feels better to me.

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: I put it back to 100 K.

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: I'll leave this for now. I'm marginally worried that an exotic char will trigger atypical code paths. In general, I think there's a rich opportunity to get more sophisticated here and figure out how to synthesize data that represents real scenarios well.

src/Microsoft.Security.Utilities.Core/CompiledHighPerformancePattern.Generated.cs
nguerrera: Again here, I will leave behavior change out of scope for this PR and file an issue.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Hmm, no, I was going too fast, but also, empty signature chars may be a footgun. Investigating.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Yeah, that would be fine here since this is an internal type. I'll leave it only because this PR is getting big.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Hmm, we have to consider `"blah blah <high-perf-match-rule1> <low-perf-match-rule2> blah blah"`. I think `yield break` would break that. I also think our testing for multiple matches in the same input is lacking right now and should be shored up.

I'm very interested in this topic, but I think it's a big one. Maybe it would have to be something like: "a lower precedence rule is not allowed to report a detection that overlaps with a detection with a higher precedence detection?".

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: Can you link the bug?

I'll leave this.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: Yes, there are similar issues on the other side with 'PreciselyClassified' including Unclassified HISv2 (but maybe your pending changes will fix that. In general, I've left out any renaming / re-categorization here from this big change even though it sort of shined a light on it for me.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: I don't like it either, I'm going to go with the majority and remove them all.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: I'm going to file one issue about improving these lists and keep these sorts of things out of this PR.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: I've zapped it. There's no actual use of Clone in the agent apart from tests. Merlyn OK'ed as well.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Actually, yes, there is no StringComparison for a search value of chars. Of course. But still, sentinel empty search values was bad. Fixed.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Done.

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: #194

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: #196

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: #196

src/Microsoft.Security.Utilities.Core/WellKnownRegexPatterns.cs
nguerrera: #196

src/Microsoft.Security.Utilities.Core/CompiledHighPerformancePattern.Generated.cs
nguerrera: #197

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_031_NuGetApiKey.cs
nguerrera: #198

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: I thought there was a standalone issue on this topic, but all I can find is a comment in #130:

> Our design today does not easily allow sharing accumulated findings between rules and ordering rules in a sensible way to provide a sort of 'chain of responsibility' design pattern for classification.

nguerrera: Also related: #95

nguerrera: Filed #199 to track this on its own.

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
michaelcfanning: Good to know this quirk! I am in code flow and it renders everything fine the `<summary>` item displays properly.

nguerrera:
Benchmarks to validate this is better across the board. UseHPS=Use High-Performance Scanner.

Command line for future reference `dotnet run -c Release -f net8.0 -- --runtimes net472 net8.0 net9.0 --join`

# Only High-Performance-Enabled Rules

| Method        | Runtime              | UseHPS | Case       | Mean           | Error        | StdDev       | Ratio |
|-------------- |--------------------- |-------------------------- |----------- |---------------:|-------------:|-------------:|------:|
| DetectSecrets | .NET 8.0             | False                     | HardPrefix |   455,115.9 us |  3,124.59 us |  2,609.18 us |  0.19 |
| DetectSecrets | .NET 9.0             | False                     | HardPrefix |   404,052.0 us |  3,072.61 us |  2,723.79 us |  0.17 |
| DetectSecrets | .NET 4.7.2 | False                     | HardPrefix | 2,403,335.8 us | 30,368.28 us | 28,406.51 us |  1.00 |
|               |                      |                           |            |                |              |              |       |
| DetectSecrets | .NET 8.0             | True                      | HardPrefix |   260,202.1 us |  1,891.96 us |  1,769.74 us |  0.43 |
| DetectSecrets | .NET 9.0             | True                      | HardPrefix |    11,268.7 us |    190.06 us |    168.49 us |  0.02 |
| DetectSecrets | .NET 4.7.2 | True                      | HardPrefix |   599,753.9 us |  4,150.52 us |  3,679.33 us |  1.00 |
|               |                      |                           |            |                |              |              |       |
| DetectSecrets | .NET 8.0             | False                     | EasyPrefix |   289,439.6 us |  3,899.65 us |  3,647.73 us |  0.20 |
| DetectSecrets | .NET 9.0             | False                     | EasyPrefix |   280,879.3 us |  3,369.92 us |  3,152.23 us |  0.19 |
| DetectSecrets | .NET 4.7.2 | False                     | EasyPrefix | 1,448,859.5 us |  4,944.74 us |  4,856.40 us |  1.00 |
|               |                      |                           |            |                |              |              |       |
| DetectSecrets | .NET 8.0             | True                      | EasyPrefix |     7,792.3 us |    144.17 us |    127.80 us |  0.19 |
| DetectSecrets | .NET 9.0             | True                      | EasyPrefix |     9,633.0 us |    190.39 us |    433.62 us |  0.24 |
| DetectSecrets | .NET 4.7.2 | True                      | EasyPrefix |    41,293.0 us |    657.93 us |    646.17 us |  1.00 |
|               |                      |                           |            |                |              |              |       |
| DetectSecrets | .NET 8.0             | False                     | NoPrefix   |     3,199.3 us |     33.59 us |     28.05 us |  0.47 |
| DetectSecrets | .NET 9.0             | False                     | NoPrefix   |     2,840.1 us |     27.45 us |     25.68 us |  0.42 |
| DetectSecrets | .NET 4.7.2 | False                     | NoPrefix   |     6,756.4 us |     49.43 us |     46.23 us |  1.00 |
|               |                      |                           |            |                |              |              |       |
| DetectSecrets | .NET 8.0             | True                      | NoPrefix   |       626.5 us |      4.95 us |      4.39 us |  0.39 |
| DetectSecrets | .NET 9.0             | True                      | NoPrefix   |       447.7 us |      1.77 us |      1.48 us |  0.28 |
| DetectSecrets | .NET 4.7.2 | True                      | NoPrefix   |     1,620.4 us |     18.01 us |     16.85 us |  1.00 |


# All PreciselyClassifiedSecurityKeys

| Method        | Runtime              | UseHPS | Case       | Mean         | Error      | StdDev     | Ratio |
|-------------- |--------------------- |-------------------------- |----------- |-------------:|-----------:|-----------:|------:|
| DetectSecrets | .NET 8.0             | False                     | HardPrefix |   644.134 ms |  4.0310 ms |  3.7706 ms |  0.15 |
| DetectSecrets | .NET 9.0             | False                     | HardPrefix |   575.450 ms |  8.5858 ms |  8.0312 ms |  0.13 |
| DetectSecrets | .NET 4.7.2 | False                     | HardPrefix | 4,285.327 ms | 23.0582 ms | 21.5686 ms |  1.00 |
|               |                      |                           |            |              |            |            |       |
| DetectSecrets | .NET 8.0             | True                      | HardPrefix |   456.673 ms |  6.9001 ms |  6.1167 ms |  0.18 |
| DetectSecrets | .NET 9.0             | True                      | HardPrefix |   151.860 ms |  1.7166 ms |  1.6057 ms |  0.06 |
| DetectSecrets | .NET 4.7.2 | True                      | HardPrefix | 2,474.910 ms | 32.8486 ms | 30.7266 ms |  1.00 |
|               |                      |                           |            |              |            |            |       |
| DetectSecrets | .NET 8.0             | False                     | EasyPrefix |   421.166 ms |  3.6499 ms |  3.4141 ms |  0.13 |
| DetectSecrets | .NET 9.0             | False                     | EasyPrefix |   444.913 ms |  7.5999 ms |  7.1089 ms |  0.13 |
| DetectSecrets | .NET 4.7.2 | False                     | EasyPrefix | 3,352.663 ms | 48.6955 ms | 45.5498 ms |  1.00 |
|               |                      |                           |            |              |            |            |       |
| DetectSecrets | .NET 8.0             | True                      | EasyPrefix |   148.337 ms |  2.0427 ms |  1.9107 ms |  0.08 |
| DetectSecrets | .NET 9.0             | True                      | EasyPrefix |   155.276 ms |  0.3187 ms |  0.2661 ms |  0.08 |
| DetectSecrets | .NET 4.7.2 | True                      | EasyPrefix | 1,927.627 ms | 19.6745 ms | 18.4035 ms |  1.00 |
|               |                      |                           |            |              |            |            |       |
| DetectSecrets | .NET 8.0             | False                     | NoPrefix   |     4.187 ms |  0.0508 ms |  0.0450 ms |  0.38 |
| DetectSecrets | .NET 9.0             | False                     | NoPrefix   |     3.740 ms |  0.0430 ms |  0.0402 ms |  0.34 |
| DetectSecrets | .NET 4.7.2 | False                     | NoPrefix   |    10.905 ms |  0.0530 ms |  0.0442 ms |  1.00 |
|               |                      |                           |            |              |            |            |       |
| DetectSecrets | .NET 8.0             | True                      | NoPrefix   |     1.519 ms |  0.0262 ms |  0.0245 ms |  0.25 |
| DetectSecrets | .NET 9.0             | True                      | NoPrefix   |     1.326 ms |  0.0162 ms |  0.0152 ms |  0.22 |
| DetectSecrets | .NET 4.7.2 | True                      | NoPrefix   |     5.969 ms |  0.0677 ms |  0.0633 ms |  1.00 |

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
michaelcfanning: Right, chars, of course. Honestly, I think in ASCII, just a bit of a dinosaur. :)

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
michaelcfanning: Yes, better.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/SEC101_565_SecretScanningSampleToken.cs
michaelcfanning: no strong feeling here, defer to code author.

src/Microsoft.Security.Utilities.Core/CompiledHighPerformancePattern.Generated.cs
michaelcfanning: Yes, you got it:

`^.{4}[0-9a-f\-]{32}(-[0-9])?`

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Sigh, I may have spoken and acted too soon here. Clone does appear to be used in agent (in a questionable but hard to change place) :(

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
michaelcfanning: Yes, thanks, let's keep talking. Related is whether we want any clear differentiation between scanning small discrete inputs vs. larger text segments with multiple findings. The whole short-circuiting thing really can't work with scanning chunks of text. We could only allow a single rule to 'grab' a specific range within the text and classify it maybe (which would prevent double firing but wouldn't afford any perf gains from dropping follow-on analysis). Eep. I'm not a big fan of it but we might actually want two discrete scanners (or perhaps two distinct Scan() methods to distinguish these cases, there are many practical differences between them.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: I looked at things in more depth and I believe there is still a path forward to keep Clone out. Let's leave it zapped in this PR with a small possiblity that I will have to revert that.

docs/ReleaseHistory.md
```diff
@@ -16,7 +16,15 @@
 - BRK, NEW: `SecretMasker.MaskSecrets` and `ISecretMasker.DetectSecrets` now accept an `Action<Detection>` callback (with a default value of `null`) to receive detections that result from the operation. This is binary-breaking only for `SecretMasker` callers and source-breaking as well for `ISecretMasker` implementations.
 - BRK: `Detection` is now sealed and immutable. All property setters and the copy constructor are removed.
 - BRK: `Detection`  no longer overrides `Equals` and `GetHashCode` nor implements `IEquatable<Detection>`. `Detection.Equals` and `Detection.GetHashCode` will therefore now be based on reference equality.
+- BRK, PRF: `IdentifiableScan` has been removed and it's functionality has been merged into `SecretMasker` which now scans the patterns it's given as fast as possible, which will be significantly faster for identifiable key patterns.
```
michaelcfanning: >it's [](http://example.com/codeflow?start=52&length=4)

This s/be `its`, best wishes, the `it's vs its` police. Sorry I missed this class D misdemeanor earlier

nguerrera: Doh! Yes.

michaelcfanning:
Not for this change, but let's consider more descriptive terms for `EasyPrefix`, et al. e.g., `ConcentratedMatches`, `SparseMatches`, or whatever.

michaelcfanning (approved):

PR #193, opened 2025-05-05
Rule versioning

michaelcfanning:
- NEW: Add `RegexPattern.CreatedVersion` and `RegexPattern.LastUpdatedVersion` to track version of rule introduction and last update to rule logic. This change has the result of emitting this information to the generated rules JSON metadata files.

This change proposes to add a mechanism for declaring the version in which a rule is first created and the version in which it was last updated. This is intended to support users that may want to incrementally enable scanning only for new/modified rules *or* want to update the package version without enabling any net new checks.

The goal here is to review/land on an API design for this and to encode the rule creation version as a first step. This entailed reviewing current release notes but also (unfortunately) examining code history, as we did not engage in the discipline of change tracking until after we'd shipped nearly a dozen packages.

All version details are now sent to the autogenerated rules metadata files.

The next step would be to have a dedicated PR that provides accurate 'last updated' version information (as gleaned from release notes).


docs/ReleaseHistory.md
```diff
@@ -194,7 +195,6 @@
 - BUG: Correct `AzureCosmosDBIdentifiableKey` rule id to `SEC101/160` (previously incorrectly listed as `SEC101/163`).
 - BUG: Correct length of `SEC101/166.AzureSearchIdentifiableQueryKey` and `SEC101/167.AzureSearchIdentifiableAdminKey` rules to 39 bytes and properly mark it as `DetectionMetadata.Identifiable`.
 - BUG: Remove `/AM7` signature + check from rust code.
-- NEW: Add `SEC101/190.AzureEventGridIdentifiableKey` check.
```
michaelcfanning, line 205: >AzureEventGridIdentifiableKey [](http://example.com/codeflow?start=23&length=29)

This rule was *not* added in 1.4.24, it was added in 1.4.20 (and in fact referenced below). #Resolved

docs/ReleaseHistory.md
```diff
@@ -237,3 +238,55 @@
 
 ## 1.4.15 - 04/16/2024
 - NEW: Implement `IdentifiableSecrets.ComputeDerivedSymmetricKey` to generate identifiable derived keys from arbitrary identifiable secrets.
+
+## 1.4.14 - 04/16/2024
```
michaelcfanning, line 250: >14 [](http://example.com/codeflow?start=7&length=2)

OK, so I went back-and-forth, we now have an explicit version and date of release for everything we shipped. I did not retroactively build release notes for release with only logic changes, I only added `RUL` tags (for every shipped rule).

I tend to find the dates helpful and to know everything in play.

 #Resolved

docs/ReleaseHistory.md
```diff
@@ -134,7 +135,7 @@
 # 1.9.1 - 11/18/2024
 - DEP: Removed dependency of the `base-62` crate in the Rust codebase, since it depended on the `failure` crate which has a known [vulnerability](https://github.com/advisories/GHSA-jq66-xh47-j9f3).
 - BUG: Fix unhandled exception raised by `CommonAnnotatedKey.TryCreate(string, out CommonAnnotatedKey)` when passed non-CASK secrets of length < 80.
-- BUG: Update `AzureEventGridIdentifiableKey` rule id to `SEC101/199` to be synced with source of the rule.
+- BUG: Correct `AzureEventGridIdentifiableKey` rule id from `SEC101/190` to `SEC101/199` to be synced with source of the rule.
```
michaelcfanning, line 138: >Correct [](http://example.com/codeflow?start=7&length=7)

I corrected the rule id elsewhere in release history where it previously referred to the incorrect version. And then I updated this release note to make it explicit that `SEC101/190` was a bogus (now fixed) rule id. #Resolved

docs/ReleaseHistory.md
```diff
@@ -218,10 +218,11 @@
 - NEW: Add `Identifiable.ComputeDerivedCommonAnnotatedKey` to generate keys derived from common annotated secrets.
 
 # 1.4.20 - 05/16/2024
+- RUL: Add `SEC101/199.AzureEventGridIdentifiableKey` detection.
```
michaelcfanning, line 221: >Add [](http://example.com/codeflow?start=7&length=3)

In fact this was a net new rule for this release, not a conversion of an existing rule like the other two cited below. #Resolved

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -57,6 +57,19 @@ protected RegexPattern()
     }
 #pragma warning restore CS8618
 
+    [DataMember(Order = 8)]
+
```
nguerrera: nit: extra blank line here. #Resolved

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
```diff
@@ -57,6 +57,19 @@ protected RegexPattern()
...
+    public virtual Version CreatedVersion
+    {
+        get
+        {
+            throw new NotImplementedException();
```
michaelcfanning: >NotImplementedException [](http://example.com/codeflow?start=22&length=23)

I first made this property abstract to ensure that all scanner rules provided it. Currently, API users can construct these patterns on the fly (and our tests do this). So, I reverted the abstract declaration and made the default implementation throw. 

Arguably, but very much not a high priority, the `RegexPattern` class could parameterize the version in the ctor to support this capability. #Resolved

nguerrera: I'd actually prefer `null` here: `public virtual Version? CreatedVersion => null`;

It's not really an error case for RegexPatterns that do not ship with the library, such as the dynamic ones created via `new RegexPattern(...)` which happens in pipeline agent, for example.

I'd then add a test that ensures that built-in patterns have a version.

It would also be worth a doc comment that this can be null if the pattern is not built-in to the library.

 #Resolved

src/Microsoft.Security.Utilities.Core/Releases.cs
```diff
@@ -0,0 +1,53 @@
...
+    public static Version Version_01_04_20 => new Version(1, 4, 20);
+    public static Version Version_01_04_19 => new Version(1, 4, 19);
+    public static Version Version_01_04_18 => new Version(1, 4, 18);
+    public static Version Version_01_04_17 => new Version(1, 4, 17);
+    public static Version Version_01_04_16 => new Version(1, 4, 16);
+    public static Version Version_01_04_15 => new Version(1, 4, 15);
+    public static Version Version_01_04_12 => new Version(1, 4, 12);
+    public static Version Version_01_04_11 => new Version(1, 4, 11);
+    public static Version Version_01_04_10 => new Version(1, 4, 10);
+    public static Version Version_01_04_02 => new Version(1, 4, 2);
```
nguerrera: Replace `=> new` with `{ get; } = new` throughout. Version is a reference type and properties shouldn't allocate a new object on every call. #Resolved

src/Microsoft.Security.Utilities.Core/Releases.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Security.Utilities;
+
+/// <summary>
+/// This class contains the version numbers of the library releases. Prior to
```
nguerrera: nit: everything after the first sentence might be more at home in `<remarks>` rather than `<summary>`. #Resolved

docs/ReleaseHistory.md
```diff
@@ -22,6 +22,7 @@
 - BRK: `WellKnownRegexPatterns.*` properties are now of stronger type`IReadOnlyList<RegexPattern>` instead of `IEnumerable<RegexPattern>`.
 - BRK: `SecretMasker.SyncObject` is now a read-only property instead of a public mutable field.
 - NEW: `SecretMasker` is now capable of finding identifiable keys without relying on delimiting characters`.
+- NEW: Add `RegexPattern.CreatedVersion` and `RegexPattern.LastUpdatedVersion` to track version of rule introduction and last update to rule logic. This change has the result of emitting this information to the generated rules JSON metadata files.
```
nguerrera: It looks like `LastUpdatedVersion` wasn't added yet? #Resolved

nguerrera: nit: "This change has the result of emitting this information to ..." is a bit wordy. Perhaps "These versions are emitted to ..."? #Resolved

src/Microsoft.Security.Utilities.Core/Releases.cs
```diff
@@ -0,0 +1,53 @@
...
+/// have added an explicit release version for every historical version where a
+/// rule was introduced. And so we have release here for version 1.4.11 (because
+/// SEC101/AadClientAppLegacyCredentials was added in this version) but we do
+/// not have an entry for 1.4.14 (because no new rules were added in this
+/// version).
+/// </summary>
+internal static class Releases
+{
+    // Current
+    public static Version Unreleased => new Version(1, 18, 0);
```
nguerrera: What is the reason for not calling this `Version_1_18_0` straight away? I'm worried that we'll forget to do that at release time. #Resolved

src/Microsoft.Security.Utilities.Core/PotentialSecurityKeys/SEC000_000_Unclassified32ByteBase64String.cs
```diff
@@ -38,14 +38,6 @@ public override IEnumerable<Detection> GetDetections(string input,
...
             string match = input.Substring(detection.Start, detection.Length);
 
-            if (!object.Equals(azure32ByteIdentifiableKeys.GetDetections(match,
-                                                                         generateSha256Hashes,
-                                                                         defaultRedactionToken,
-                                                                         regexEngine).FirstOrDefault(), objB: default))
-            {
-                continue;
-            }
-
```
nguerrera: Hmm, I don't see this mentioned in PR description or release notes.

We're giving up on avoiding double-firing in a cumbersome way? This was actually my motivating example in #199.

If this is intended to stay out, then I think the whole override can be removed and Azure32ByteIdentifiableKeys can be deleted.

(Ditto all of the above for Azure64Byte.) #Resolved

src/Microsoft.Security.Utilities.Core/PotentialSecurityKeys/SEC000_001_Unclassified64ByteBase64String.cs
```diff
@@ -38,14 +39,6 @@ public override IEnumerable<Detection> GetDetections(string input,
...
             string match = input.Substring(detection.Start, detection.Length);
 
-            if (!Equals(azure64ByteIdentifiableKeys.GetDetections(match,
-                                                                         generateSha256Hashes,
-                                                                         defaultRedactionToken,
-                                                                         regexEngine).FirstOrDefault(), objB: default))
-            {
-                continue;
-            }
-
```
nguerrera: (Same here as Azure32Byte.) #Resolved

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
nguerrera: See my suggestion to make this null by default. I don't think a constructor arg works. If you're newing up a RegexPattern outside the library, there just isn't a library version you were created in.

docs/ReleaseHistory.md
nguerrera: Nevermind, I'm just wrong here. :)

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
nguerrera: (Fixed while I was still reviewing)

src/Microsoft.Security.Utilities.Core/Releases.cs
michaelcfanning: The reason I had in mind is that the specific version implied by `Unreleased` isn't really known until we finish all the work. e.g., `1.18` could turn into `2.0` depending on what happens in the next check-in.

As we prepare for release, what we'd do is perform an IDE rename of `Unreleased` to the final version and then put in a new prospective `Unreleased` value.

I thought this was easiest. But maybe it's more theater than we need. If we take your suggestion, we just use `1_18` as long as it's relevant and then rename it on accepting an ureleased major breaking change. OK, so I'll take your note. :)

src/Microsoft.Security.Utilities.Core/PotentialSecurityKeys/SEC000_000_Unclassified32ByteBase64String.cs
michaelcfanning: No, sorry, this is a good catch, bad code removal! reverting

docs/ReleaseHistory.md
michaelcfanning: nice yes.

src/Microsoft.Security.Utilities.Core/RegexPattern.cs
michaelcfanning: Perfect, yes.

src/Microsoft.Security.Utilities.Core/Releases.cs
```diff
@@ -0,0 +1,52 @@
...
+/// release history and we have not back-filled release notes for all code
+/// changes. We have added an explicit release version for every historical
+/// version where a rule was introduced. And so we have release here for version
+/// 1.4.11 (because SEC101/AadClientAppLegacyCredentials was added in this
+/// version) but we do not have an entry for 1.4.14 (because no new rules were
+/// added in this version).
+/// </remarks>
+internal static class Releases
+{
+    public static Version Version_01_18_00 => new Version(1, 18, 0);
```
nguerrera: This one is still a property that allocates.

src/Microsoft.Security.Utilities.Core/Releases.cs
```diff
@@ -6,48 +6,47 @@
...
-    public static Version Version_01_04_18 => new Version(1, 4, 18);
-    public static Version Version_01_04_17 => new Version(1, 4, 17);
-    public static Version Version_01_04_16 => new Version(1, 4, 16);
-    public static Version Version_01_04_15 => new Version(1, 4, 15);
-    public static Version Version_01_04_12 => new Version(1, 4, 12);
-    public static Version Version_01_04_11 => new Version(1, 4, 11);
-    public static Version Version_01_04_10 => new Version(1, 4, 10);
-    public static Version Version_01_04_02 => new Version(1, 4, 2);
+    public static Version Version_01_18_00 => new Version(1, 18, 0);
+    public static Version Version_01_17_00 = new Version(1, 17, 0);
```
nguerrera, line 23: These shouldn't be mutable. Either `readonly` or `{ get; } =` would be fine alternatives to fix that.

src/Microsoft.Security.Utilities.Core/Releases.cs
nguerrera: (Fixed while I was reviewing)

nguerrera (approved):

PR #202, opened 2025-05-07
Granular legacy cask rules

michaelcfanning:
- RUL: Add `SEC101/177.SqlPrivateDefaultCloudSALegacyCommonAnnotatedSecurityKey`, `SEC101/197.AzureAppConfigurationLegacyCommonAnnotatedSecurityKey`, `SEC101/198.AzureFluidRelayLegacyCommonAnnotatedSecurityKey`, `SEC101/199.AzureEventGridLegacyCommonAnnotatedSecurityKey`, `SEC101/201.AzureDevOpsLegacyCommonAnnotatedSecurityKeyPat`, `SEC101/202.AzureMixedRealityLegacyCommonAnnotatedSecurityKeyPat`, `SEC101/203.AzureMapsLegacyCommonAnnotatedSecurityKey`, `SEC101/204.AzureCommunicationServicesLegacyCommonAnnotatedSecurityKey`, `SEC101/205.AzureAIServicesLegacyCommonAnnotatedSecurityKey`, `SEC101/206.AzureOpenAILegacyCommonAnnotatedSecurityKey`, `SEC101/207.AzureAnomalyDetectorEELegacyCommonAnnotatedSecurityKey`, `SEC101/208.AzureAnomalyDetectorLegacyCommonAnnotatedSecurityKey`, `SEC101/209.AzureCognitiveServicesLegacyCommonAnnotatedSecurityKey`, `SEC101/210.AzureComputerVisionLegacyCommonAnnotatedSecurityKey`, `SEC101/211.AzureContentModeratorLegacyCommonAnnotatedSecurityKey`, `SEC101/212.AzureContentSafetyLegacyCommonAnnotatedSecurityKey`, `SEC101/213.AzureCustomVisionPredictionLegacyCommonAnnotatedSecurityKey`, `SEC101/214.AzureCustomVisionTrainingLegacyCommonAnnotatedSecurityKey`, `SEC101/215.AzureFaceLegacyCommonAnnotatedSecurityKey`, `SEC101/216.AzureFormRecognizerLegacyCommonAnnotatedSecurityKey`, `SEC101/217.AzureHealthDecisionSupportLegacyCommonAnnotatedSecurityKey`, `SEC101/218.AzureHealthInsightsLegacyCommonAnnotatedSecurityKey`, `SEC101/219.AzureImmersiveReaderLegacyCommonAnnotatedSecurityKey`, `SEC101/220.AzureInternalAllInOneLegacyCommonAnnotatedSecurityKey`, `SEC101/221.AzureKnowledgeLegacyCommonAnnotatedSecurityKey`, `SEC101/222.AzureLuisAuthoringLegacyCommonAnnotatedSecurityKey`, `SEC101/223.AzureLuisLegacyCommonAnnotatedSecurityKey`, `SEC101/224.AzureMetricsAdvisorLegacyCommonAnnotatedSecurityKey`, `SEC101/225.AzurePersonalizerLegacyCommonAnnotatedSecurityKey`, `SEC101/226.AzureQnAMakerLegacyCommonAnnotatedSecurityKey`, `SEC101/227.AzureQnAMakerv2LegacyCommonAnnotatedSecurityKey`, `SEC101/228.AzureSpeakerRecognitionLegacyCommonAnnotatedSecurityKey`, `SEC101/229.AzureSpeechServicesLegacyCommonAnnotatedSecurityKey`, `SEC101/230.AzureSpeechTranslationLegacyCommonAnnotatedSecurityKey`, `SEC101/231.AzureTextAnalyticsLegacyCommonAnnotatedSecurityKey`, `SEC101/232.AzureTextTranslationLegacyCommonAnnotatedSecurityKey`, `SEC101/233.AzureDummyLegacyCommonAnnotatedSecurityKey`, `SEC101/234.AzureTranscriptionIntelligenceLegacyCommonAnnotatedSecurityKey`, `SEC101/235.AzureVideoIntelligenceLegacyCommonAnnotatedSecurityKey`, `SEC101/236.AzureBingAutosuggestLegacyCommonAnnotatedSecurityKey`, `SEC101/237.AzureBingAutosuggestv7LegacyCommonAnnotatedSecurityKey`, `SEC101/238.AzureBingCustomSearchLegacyCommonAnnotatedSecurityKey`, `SEC101/239.AzureBingCustomVisualSearchLegacyCommonAnnotatedSecurityKey`, `SEC101/240.AzureBingEntitySearchLegacyCommonAnnotatedSecurityKey`, `SEC101/241.AzureBingSearchLegacyCommonAnnotatedSecurityKey`, `SEC101/242.AzureBingSearchv7LegacyCommonAnnotatedSecurityKey`, `SEC101/243.AzureBingSpeechLegacyCommonAnnotatedSecurityKey`, `SEC101/244.AzureBingSpellCheckLegacyCommonAnnotatedSecurityKey`, and `SEC101/245.AzureBingSpellCheckv7LegacyCommonAnnotatedSecurityKey` detections.  

This change provides granular detections for all legacy CASK keys.

While also attempting to preserve the caching behaviors for the high-performance scanning capability.

Benchmark data:

### MAIN : ~870 examples, 26 patterns
| Method        | Case       | Mean         | Error       | StdDev      |
|-------------- |----------- |-------------:|------------:|------------:|
| DetectSecrets | HardPrefix | 405,378.1 us | 3,819.36 us | 3,385.77 us |
| DetectSecrets | EasyPrefix |  14,630.4 us |   145.16 us |   128.68 us |
| DetectSecrets | NoPrefix   |     937.6 us |    16.91 us |    15.82 us |

### NEW RULES: limited to 870 examples, 79 patterns, improved `GetMatchIdAndName` 
| Method        | Case       | Mean       | Error     | StdDev    |
|-------------- |----------- |-----------:|----------:|----------:|
| DetectSecrets | HardPrefix | 404.912 ms | 6.9134 ms | 6.4668 ms |
| DetectSecrets | EasyPrefix |  15.442 ms | 0.2890 ms | 0.3092 ms |
| DetectSecrets | NoPrefix   |   1.271 ms | 0.0229 ms | 0.0215 ms |


GeneratedRegexPatterns/HighConfidenceSecurityModels.json
```diff
@@ -289,6 +301,582 @@
...
+    "Pattern": "(^|[^abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890])(?P<refine>[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{52}JQQJ99[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890][A-L][abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{12}AAAbACOG[abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890]{4})",
+    "Signatures": [
+      "JQQJ"
+    ],
+    "CreatedVersion": "1.18.0",
+    "LastUpdatedVersion": "1.18.0"
+  },
+  {
+    "Id": "SEC101/233",
+    "Name": "AzureDummyLegacyCommonAnnotatedSecurityKey",
```
michaelcfanning, line 726: >AzureDummyLegacyCommonAnnotatedSecurityKey [](http://example.com/codeflow?start=13&length=42)

Cognitive services has reserved a key kind for the body of granular providers built on their platform. Some of these providers are for test purposes. Others have been deprecated or are no longer in use. We could investigate and attempt a more tuned inventory, all these checks certainly have test impact (notably! the net452 test execution is now double other CLR versions). OTOH, if the high performance regex caching is doing what it's supposed to, the overly broad rules definitions shouldn't have practical impact executing our API, yes? #ByDesign

src/Microsoft.Security.Utilities.Cli/ScanOptions.cs
```diff
@@ -16,7 +16,7 @@ public class ScanOptions
         [Option(
             's',
             "string-input",
-            HelpText = "A path to a file to scan for identifiable secrets.")]
+            HelpText = "A hard-coded input to scan for identifiable secrets.")]
```
michaelcfanning, line 19: >hard-coded input [](http://example.com/codeflow?start=26&length=16)

I used this command for testing and was compelled to fix this. I am not documenting it in the release notes. For one thing, it's a very bad practice to provide a command-line argument that contains arbitrary untrusted input we expect might contain a secret. 

https://github.com/microsoft/security-utilities/issues/203 #Resolved

KalleOlaviNiemitalo:
Unfortunate that the "legacy" CASK format specification does not have a version number. #Closed

michaelcfanning:
`JQQJ` in the relevant location is effectively the version. :) More recent CASK (which allows for some key length variability) puts the fixed signature in a new location, and also changes the signature to `QJJQ`. It's unfortunate that we didn't build in a versioning mechanism to the older format, however, your points stands. We will try to 'fix' this in the open standard CASK.

---
In reply to: [2863836126](https://github.com/microsoft/security-utilities/pull/202#issuecomment-2863836126) [](http://example.com/codeflow?ancestors=2863836126)

KalleOlaviNiemitalo:
I don't mean a version marker encoded within the CASK string.  I mean a version of the CASK specification so that you'd be able to call it something like "CASK 0.1" or "CASK 2024", rather than "legacy". #Pending

src/Microsoft.Security.Utilities.Core/LegacyCaskProviderSignatures.cs
```diff
@@ -0,0 +1,121 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities;
+
+public static class LegacyCaskProviderSignatures
+{
+    public static ISet<string> All = new HashSet<string>
```
nguerrera:
```suggestion
    public static ISet<string> All { get; } = new HashSet<string>
```
 #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/AzureCognitiveServices.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Security.Utilities;
+
+public enum AzureCognitiveServices
```
nguerrera: Non-flags enum should not be plural.

```suggestion
public enum AzureCognitiveService
``` #Resolved

File: src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/CognitiveServiceLegacyCommonAnnotatedSecurityKey.cs [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+        AzureCognitiveService = service;
+    }
+
+    protected override string ProviderSignature => LegacyCaskProviderSignatures.AzureCognitiveServices;
+
+    protected AzureCognitiveServices AzureCognitiveService { get; }
+
+    public override Tuple<string, string> GetMatchIdAndName(string match)
+    {
+        if (match.Length == 88) // comment or constant please
```
nguerrera: Why does this not have a corresponding change in general regex pattern from base? It looks like base also expects fixed length. Should this check be moved to base and this override deleted? #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/LegacyCommonAnnotatedSecurityAccessKey.cs
```diff
@@ -0,0 +1,98 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Security.Utilities
+{
+    public abstract class LegacyCommonAnnotatedSecurityAccessKey : UnclassifiedLegacyCommonAnnotatedSecurityKey
```
nguerrera, line 9: This inheritance relationship is confusing and questionable. We are saying that each classified key below this *is* an unclassified key.
 
I would make the Unclassified rule a leaf node. #Pending

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/LegacyCommonAnnotatedSecurityAccessKey.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            ProviderData = providerData;
+            DetectionMetadata = DetectionMetadata.Identifiable;
+            Pattern = $"{WellKnownRegexPatterns.PrefixBase62}(?P<refine>[{WellKnownRegexPatterns.Base62}]{{52}}JQQJ99[{WellKnownRegexPatterns.Base62}][A-L]{PlatformData}{ProviderData}{ProviderSignature}[{WellKnownRegexPatterns.Base62}]{{4}})";
+            Signatures = new HashSet<string>([LegacyCaskSignature]);
+        }
+
+        public override Tuple<string, string> GetMatchIdAndName(string match)
+        {
+#if DEBUG
```
nguerrera: I just realized this is probably why buildandtest is reasonably fast but run all tests in IDE is very slow. 

Can we delete this? Maybe we can include a comment on why the math is good enough for high confidence without a checksum validation? #Resolved

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/AzureCognitiveServices.cs
```diff
@@ -0,0 +1,49 @@
...
+    AzureHealthInsights = 13,
+    AzureImmersiveReader = 14,
+    AzureInternalAllInOne = 15,
+    AzureKnowledge = 16,
+    AzureLuisAuthoring = 17,
+    AzureLuis = 18,
+    AzureMetricsAdvisor = 19,
+    AzurePersonalizer = 20,
+    AzureQnAMaker = 21,
+    AzureQnAMakerv2 = 22,
```
nguerrera: Since there are others like this, I assume it's intentional, but should the casing be `V2` with a capital V? #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -119,16 +119,34 @@ private static int RoundUpToMultipleOf(int value, int multiple)
...
+            }
+
+            while (examples.Count < examplesCount)
+            {
+                foreach (IEnumerator<string> exampleGenerator in exampleGenerators)
                 {
-                    examples.Add($"{prefix} {example}");
+                    if (exampleGenerator.Current != null)
+                    {
+                        examples.Add(prefix + exampleGenerator.Current);
```
nguerrera: I think you will need the space as before when high-perf scanner is not used because the general regexes rely on delimeters. #Resolved

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
nguerrera: I think we're OK on perf now, but the name is bugging me. Can we use something other than 'dummy' here?

nguerrera:
On Legacy name, I'm a bit concerned that we lose the idea that a "legacy key" is a key that should no longer be issued. There's no urgency for these to adopt non-legacy CASK.

I've also found `IdentifiableKey` vs `(Legacy)CommonAnnotatedSecurityKey` confusing. The latter are identifiable as well. 

These are also making really long class names.

I'd be very tempted to use these names:
* `Identifiable` -> `HisV1`
* `LegacyCommonAnnotatedSecurityKey` -> `HisV2`
* `CommonAnnotatedSecurityKey` -> `Cask`

I think this matches how we commonly talk about them among ourselves. We'd be bending the rule on when to use acronyms, but I think that's acceptable here. If you're not an expert on the distinction between these, I don't think you care about this portion of a rule moniker.

 Just thinking out loud. If we do rename anything, that should be a separate change.
 #Pending

nguerrera:
Can you update the benchmark data in the description? #Resolved

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
michaelcfanning: Nuts, LLM converted this and I didn't notice, thank you.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/AzureCognitiveServices.cs
michaelcfanning: Thank you, FxCop engineer emeritus. I suppose you didn't comment on the absence of a zero `None` value because we are non-flags and, in fact, put the most common/default value in this position.

michaelcfanning:
OK, got it, yes. Let's think about this. Nick has thoughts as well I see.

---
In reply to: [2863990858](https://github.com/microsoft/security-utilities/pull/202#issuecomment-2863990858) [](http://example.com/codeflow?ancestors=2863990858)

michaelcfanning:
OK, sure. Hm. `HisV1` and `HisV2` are really not great as names, we've skipped right to obscure acronyms with no one really understanding the terms to start with. Other ideas: `Identfiable`, `LegacyCask`, `Cask`. Alternately! LegacyIdentifiable, `Identifiable`, and CASK. This latter formulation is honestly a bit more accurate, HIS v1 was the first thing we called 'identifiable' and the next one was a revision on it. We have no obviously good options here, imo. Let's keep talking. [https://github.com/microsoft/security-utilities/issues/205](https://github.com/microsoft/security-utilities/issues/205)





---
In reply to: [2867678801](https://github.com/microsoft/security-utilities/pull/202#issuecomment-2867678801) [](http://example.com/codeflow?ancestors=2867678801)

GeneratedRegexPatterns/HighConfidenceSecurityModels.json
michaelcfanning: That's the actual searchable name. :) Which argues for keeping it. I don't have a precise sense of what purpose this provider has, so I hesitate to finesse the language. I could ask about it if you feel strongly. ? What's your SEV on this, sir?

michaelcfanning:
Done and for the record planned to do so. :) New data looks good. Nice work trimming the time, sir, thanks.

---
In reply to: [2867686732](https://github.com/microsoft/security-utilities/pull/202#issuecomment-2867686732) [](http://example.com/codeflow?ancestors=2867686732)

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/CognitiveServiceLegacyCommonAnnotatedSecurityKey.cs
michaelcfanning: Good eye. First, I've updated the comment. Second, the formal definition of an access key allows for both the short form and the long form, so the base class regex was broken. now fixed. thanks for the note.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/LegacyCommonAnnotatedSecurityAccessKey.cs
michaelcfanning: sure thing! honestly, now looking at this, it's a little dumb. probably what we wanted here was a debug assert for this condition, not some conditionally compiled check that was only in play for debug builds.

michaelcfanning: Also, we ran this code twice, once in the unclassified base and then the derived. Ouch. Now zapped and it helps a lot.

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/LegacyCommonAnnotatedSecurityAccessKey.cs
michaelcfanning: 100%. [https://github.com/microsoft/security-utilities/issues/206](https://github.com/microsoft/security-utilities/issues/206)

src/Microsoft.Security.Utilities.Core/PreciselyClassifiedSecurityKeys/AzureCognitiveServices.cs
michaelcfanning: Yes, this casing mistake was picked up from provider project, there's no problem correcting it, done.

nguerrera (approved):

PR #208, opened 2025-05-13
Multi-threading and minimum secret length improvements

nguerrera:
1. Seal SecretMasker and remove all visible mutable state. It is basically impossible to reason about thread safety without this.
2. Support recursion on ReaderWriterLock. Allows a wrapping caller with additional state to broaden the scope of a lock.
3. Fix race condition in elapsed masking time computation.
4. Count time in MaskSecrets exclusive of DetectSecrets in masking time computation.
5. Respect MinimumSecretLength for regex-based (non-literal) matches.
6. Make lengthening and re-shortening of MinimumSecretLength non-destructive.
7. Simplify AddValue and AddLiteralEncoder, fix race conditions, and remove corresponding "assumes no more than one thread" comments.
8. Stop using iterators to implement methods that take locks. This can lead to deadlock if enumerator is not disposed or just take a lock for a long time if a thread is doing a lot of work that needn't be synchronized with each element before disposing at the end.
9. Remove unused MinimumSecretLengthCeiling.


michaelcfanning:
We need to update the release notes, yes? for the `BRK` at least #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -173,175 +176,124 @@ public string MaskSecrets(string input, Action<Detection>? detectionAction = nul
...
     /// Gets or sets the minimum allowable size of a string that's a candidate for masking.
     /// </summary>
-    public virtual int MinimumSecretLength { get; set; }
-
-    /// <summary>
-    /// This implementation assumes no more than one thread is adding regexes, values, or encoders at any given time.
-    /// </summary>
-    [ExcludeFromCodeCoverage]
-    public void AddValue(string value)
+    public int MinimumSecretLength
```
michaelcfanning: >MinimumSecretLength [](http://example.com/codeflow?start=15&length=19)

too bad we can't just rip this out. why can't the code that populates the masker handle this (i.e., not populate as appropriate when they do that)? i know you have a good answer so mostly just expressing a feeling here. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -38,55 +38,46 @@ public SecretMasker() : this(default, default, default, default, default)
...
                         string? defaultRegexRedactionToken = null,
                         string? defaultLiteralRedactionToken = null)
     {
-        _disposed = false;
-
-        RegexPatterns = regexSecrets != null
-            ? new HashSet<RegexPattern>(regexSecrets)
-            : new HashSet<RegexPattern>();
-
+        _regexPatterns = new HashSet<RegexPattern>(regexSecrets ?? []);
```
michaelcfanning, line 55: >[] [](http://example.com/codeflow?start=67&length=2)

NIT: maybe simply set default arg to `[]` instead of null ? #Resolved

nguerrera:
> We need to update the release notes, yes? for the `BRK` at least

Yes. I also want to add some tests. Both reasons why I haven't taken this out of draft yet.  #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: I wish that were possible but `[]` is not a compile-time constant and therefore can't be a default arg.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -173,175 +176,124 @@ public string MaskSecrets(string input, Action<Detection>? detectionAction = nul
...
-        {
-            if (SyncObject.IsReadLockHeld)
+            finally
             {
-                SyncObject.ExitReadLock();
+                SyncObject.ExitWriteLock();
             }
         }
+    }
+    private int _minimumSecretLength;
```
michaelcfanning: >private [](http://example.com/codeflow?start=4&length=7)

NIT: whitespace before this field declaration. do you prefer to co-locate property backing fields close to accessor declarations? we could also group all privates somewhere. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -173,175 +176,124 @@ public string MaskSecrets(string input, Action<Detection>? detectionAction = nul
...
 
-        // Compute the encoded values.
-        foreach (LiteralEncoder literalEncoder in literalEncoders)
+    public void AddValue(string value)
+    {
+        if (string.IsNullOrEmpty(value))
```
michaelcfanning, line 181: >IsNullOrEmpty [](http://example.com/codeflow?start=19&length=13)

realistically, IsNullOrWhitespace would make sense, no secret literal can comprise all whitespace. but also, why would you pass this data? #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: So I was tempted to group all the private fields together at the top, that's my usual way, but then I didn't want to make a bigger diff than necessary. But if you'd prefer that too, then I can go ahead and do it. :)

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -173,175 +176,124 @@ public string MaskSecrets(string input, Action<Detection>? detectionAction = nul
...
-                SyncObject.ExitWriteLock();
-            }
+            Interlocked.Add(ref _elapsedMaskingTicks, stopwatch.ElapsedTicks);
         }
     }
 
-    public IEnumerable<Detection> DetectSecrets(string input)
+    private List<Detection> DetectSecretsCore(string input)
     {
         if (string.IsNullOrEmpty(input))
```
michaelcfanning: >IsNullOrEmpty [](http://example.com/codeflow?start=19&length=13)

same NIT, is IsNullOrWhitespace the right thing here? #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -350,26 +302,31 @@ public IEnumerable<Detection> DetectSecrets(string input)
...
+                    {
+                        detections.Add(detection);
+                    }
                 }
             }
 
-            foreach (SecretLiteral secretLiteral in EncodedSecretLiterals)
+            foreach (SecretLiteral secretLiteral in _encodedSecretLiterals)
             {
+                if (secretLiteral.Value.Length < MinimumSecretLength)
```
michaelcfanning: >MinimumSecretLength [](http://example.com/codeflow?start=49&length=19)

Every regex should be enforcing minimal lengths not this API. Sorry for raising this point again. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -38,55 +38,46 @@ public SecretMasker() : this(default, default, default, default, default)
...
     /// <summary>
-    /// Gets the total time in ticks spent masking content for the lifetime of this masker instance.
+    /// Gets the total time spent masking content for the lifetime of this masker instance.
     /// </summary>
-    public long ElapsedMaskingTime { get; private set; }
+    public TimeSpan ElapsedMaskingTime => TimeSpan.FromTicks(_elapsedMaskingTicks);
+    private long _elapsedMaskingTicks;
 
     /// <summary>
     /// This implementation assumes no more than one thread is adding regexes, values, or encoders at any given time.
```
michaelcfanning: >This implementation assumes no more than one thread is adding regexe [](http://example.com/codeflow?start=8&length=68)

i don't think this comment is correct any longer due to changes you have made, yes? #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Security.Utilities;
 /// <returns>The escaped or encoded literal.</returns>
 public delegate string LiteralEncoder(string literal);
 
-public class SecretMasker : ISecretMasker
+public sealed class SecretMasker : ISecretMasker
 {
     private readonly IRegexEngine _regexEngine;
```
michaelcfanning, line 24: >private [](http://example.com/codeflow?start=4&length=7)

we've got privates, both readonly and not sprinkled throughout this class, can we just group them somewhere? SEV4 #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Yes, too bad!
 
I've simplified it a lot and removed the need for RemoveShortSecretsFromDictionary in the agent. But the existence of that method over there shows the problem (that was solved awkwardly) where they set minimum secret length is set after values have been added.

The agent also had really weird code to remove based regex length rather than regex *match* length. 

Removing things was also broken in the theoretical case of shortening the min length (removes) than re-lengthening (doesn't re-add).

I've simplified things so that the integer is the only mutable state and we respect it always (and do sanely in the case of regexes).

This landing point is not how I would have designed things from scratch, but still much easier to maintain between here and the agent than where we were before this change, so I'll take it. :)

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -494,11 +436,9 @@ internal void DisableHighPerformanceScannerForTests()
...
-    public HashSet<SecretLiteral> EncodedSecretLiterals { get; }
-    public HashSet<SecretLiteral> ExplicitlyAddedSecretLiterals { get; }
-
-    public ReaderWriterLockSlim SyncObject { get; } = new(LockRecursionPolicy.NoRecursion);
+    private readonly HashSet<LiteralEncoder> _literalEncoders;
+    private readonly HashSet<SecretLiteral> _encodedSecretLiterals;
+    private readonly HashSet<SecretLiteral> _explicitlyAddedSecretLiterals;
 
-    private bool _disposed;
+    public ReaderWriterLockSlim SyncObject { get; } = new(LockRecursionPolicy.SupportsRecursion);
```
michaelcfanning, line 429: >new [](http://example.com/codeflow?start=54&length=3)

I'm surprised style checker doesn't want a space here or are we treating `new` as a sort of function name? IIRC you can't win with this simplified construction pattern, one or the other of some style issue objects (but maybe not if we're configured to prevent that). SEV5 #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -38,55 +38,46 @@ public SecretMasker() : this(default, default, default, default, default)
...
-    // We don't permit secrets great than 5 characters in length to be
-    // skipped at masking time. The secrets that will be ignored when
-    // masking will N - 1 of this property value.
-    public static int MinimumSecretLengthCeiling { get; set; }
-
     [ThreadStatic]
     private static StringBuilder? s_stringBuilder;
 
-    public virtual string DefaultRegexRedactionToken { get; set; }
+    public string DefaultRegexRedactionToken { get; }
```
michaelcfanning, line 70: >DefaultRegexRedactionToken [](http://example.com/codeflow?start=18&length=26)

These are no longer 'default' if they are instance data. Instead, they are simply the configured `RegexRedactionToken` etc. #Pending

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -173,175 +176,124 @@ public string MaskSecrets(string input, Action<Detection>? detectionAction = nul
...
             {
-                foreach (HighPerformanceDetection highPerformanceDetection in _highPerformanceScanner.Scan(input))
+                foreach (HighPerformanceDetection highPerformanceDetection in highPerformanceDetections)
                 {
                     string found = input.Substring(highPerformanceDetection.Start, highPerformanceDetection.Length);
                     foreach (RegexPattern pattern in _highPerformanceSignatureToPatternsMap![highPerformanceDetection.Signature])
                     {
                         Detection? detection = FinalizeHighPerformanceDetection(highPerformanceDetection, found, pattern);
-                        if (detection != null)
+                        if (detection != null && detection.Length >= MinimumSecretLength)
```
michaelcfanning: >MinimumSecretLength [](http://example.com/codeflow?start=69&length=19)

See, this confuses me. I think the point of `MinimumSecretLength` is (or should be) to allow short circuiting of detection and masking operations if your input data is so short that there can't be a secret there. It doesn't make sense to say, we found something, but it's too short to report. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
michaelcfanning: >IsNullOrEmpty [](http://example.com/codeflow?start=19&length=13)

if you input is < minimum secret length, there's no point in taking a lock and attempting any scanning #Resolved

michaelcfanning:
>         foreach (var pattern in patterns)

i don't like these uses of var where the type is entirely elided. SEV3 #Resolved

---
Refers to: src/Microsoft.Security.Utilities.Core/SecretMasker.cs:392 in 4c7d0c0. [](commit_id = 4c7d0c05c1408813006157692a4dc724c1f9c16f, deletion_comment = False)

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: This is how I've always seen `target-typed new` formatted so I believe it's by design that the formatter wants it this way. See, for example, the formatting in the docs. 

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/target-typed-new

michaelcfanning (approved):
:shipit:

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Yes, I agree and I'll do it. I was being overly conservative in trying to minimize the diff, but also strongly wanted to do this!

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Correct. I thought I got all of these. Will fix.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Yes, this minimum length feature sucks, see other thread, lol. 

But as long as we have both of these:

1. Public API that accepts arbitrary regex.
2. Public API that accepts arbitrary minimum length.

I think it follows that we have should have this check.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: See other thread. We agree and this will be done. :)

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: There's a real enough difference between these that I'm not inclined to change this behavior right now in this PR.

Masking "   " is weird but well-defined.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: I don't think that is a compatible interpretation with how the agent works and I also don't think that interpretation matches the name "minimum secret length".

nguerrera: We talked about this all at length offline. I am keeping what I have AND adding an additional optimization for short input.

But thinking about this purely in terms of short input data doesn't work. You can have long input with short secrets inside it.

nguerrera:
> > ```
> >     foreach (var pattern in patterns)
> > ```
> 
> i don't like these uses of var where the type is entirely elided. SEV3
> 

These aren't part of this change.

I will update the formatter separately to enforce this.

It will make me a bit sad since I DO like this, but I'll do it.

EDIT: #211 
 #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: These were already instance data. Also, the regex redaction token is still a default. If c3ids are generated than it will prefer ID:c3id over this.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Resolving as dupe of other thread.

nguerrera:
Still working on tests, will add some plus release notes before taking out of draft.

docs/ReleaseHistory.md
```diff
@@ -12,6 +12,15 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- BRK, NEW: A shared `SecretMasker` instance is now fully safe to use from multiple threads. To establish this, it is now sealed and properties that exposed unsynchronized mutable state have been made read-only or removed. The following properties are now read-only: `DefaultRegexRedactionToken`, `DefaultLiteralRedactionToken`. The following properties that returned mutable collections are now removed: `RegexPatterns`, `EncodedSecretLiterals`, `ExplicitlyAddedSecretLiterals`, `LiteralEncoders`.`EncodedSecretLiterals`, `ExplicitlyAddedSecretLiterals`. The limitation that `AddValue` and `AddLiteralEncoder` were only thread safe if there was only one thread adding values is also removed.
+- BRK, NEW: For clarity, `SecretMasker.ElapsedTime` now returns a `TimeSpan` instead of a `long` tick count.
+- BRK, BUG: `SecretMasker.MinimumSecretLengthCeiling` is removed. This property had no effect previously.
```
michaelcfanning, line 16: >BUG [](http://example.com/codeflow?start=7&length=3)

SEV6: BRK, BUG arguably sorts before BRK, NEW. Your friend, low severity code reviewer. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
michaelcfanning: OK, let's change these names in a future change.

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -114,20 +105,32 @@ public virtual void AddRegex(RegexPattern regexSecret)
...
+    {
+        if (string.IsNullOrEmpty(input))
         {
             return string.Empty;
         }
 
-        var enumerableDetections = DetectSecrets(input);
+        List<Detection> detections = DetectSecretsCore(input);
 
         // Short-circuit if nothing to replace.
```
michaelcfanning: >// Short-circuit if nothing to replace [](http://example.com/codeflow?start=8&length=38)

SEV5: low value comment. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -171,177 +174,125 @@ public string MaskSecrets(string input, Action<Detection>? detectionAction = nul
...
+            return;
         }
 
-        // Write section.
+        SyncObject.EnterWriteLock();
         try
         {
-            SyncObject.EnterWriteLock();
+            SecretLiteral literal = new SecretLiteral(value);
+            if (!_explicitlyAddedSecretLiterals.Add(literal))
```
michaelcfanning, line 190: SEV4: strictly speaking this construction is thread-safe and be completed before we take the lock. #Resolved

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -171,177 +174,125 @@ public string MaskSecrets(string input, Action<Detection>? detectionAction = nul
...
+            return [];
         }
 
-        // Read section.
+        SyncObject.EnterReadLock();
         try
         {
-            SyncObject.EnterReadLock();
-            var stopwatch = Stopwatch.StartNew();
+            var detections = new List<Detection>();
```
michaelcfanning: >List [](http://example.com/codeflow?start=33&length=4)

Gee, we sure have a lot of code within this lock. I look forward to talking to you about how we might improve this moving forward. Basically by controlling the configuration of regexes/literals/etc.

And then we only have the pesky minimum secret length thing to worry about. If it ever comes to it, though, maybe we could produce all the detections up front, and then only take the lock to apply the minimum secret length filter. 

Or you know what, just copy this value upfront, so that any multithreaded change around that value only is applied to any scans initiated *after* the change.
 #Resolved

michaelcfanning (changes requested):
:clock1:

michaelcfanning (approved):
:shipit:

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
nguerrera: Yes, there's a lot of opportunity here. One idea I have is to make the entire backing state a persistent data structure. Then we could elimintate the lock. This would require adding a System.Collections.Immutable dependency for .NET Standard/Framework, but dependencies are coming soon anyway for Span. It will also require some measurement on the perf tradeoff for single threaded vs multi-threaded use. We would also, IMHO need to push all the telemetry collection of consumers into a feature of the library. Currently pipeline agent PR relies on being able to use SyncObject. I decided this was a good stopping point where we basically have the same synchronization as status quo with the API more locked down to make future improvement easier.

nguerrera: Hmm, but I put a write lock on the minimum secret length updating and then moved a read out of the read lock in response to PR feedback. I will go ahead and make minimum secret length lock-free in the way you suggest now and add a test.

michaelcfanning (changes requested):
:clock1:

michaelcfanning (approved):
:shipit:

PR #211, opened 2025-05-14
Enforce and automatically format `var` style preferences

nguerrera:
* built-in types -> explicit type
* apparent type -> var
* non-apparent type -> explicit type

This isn't my personal preference, but it is the prevailing style and PR feedback tends to reflect that so I'm automating that.


src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -91,8 +91,8 @@ private string GeneratePrefix(int length, bool random)
 
             if (random)
             {
-                var byteCount = Base64CharsToBytes(length);
-                var data = new byte[byteCount];
+                int byteCount = Base64CharsToBytes(length);
+                byte[] data = new byte[byteCount];
```
michaelcfanning, line 95: > byte[]  [](http://example.com/codeflow?start=15&length=8)

var is fine here. why doesn't the compiler see this type as apparent? [EDITED: I see elsewhere in code that array allocations aren't regarded as 'type apparent'. ok. disagree but ok]

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
```diff
@@ -91,8 +91,8 @@ private string GeneratePrefix(int length, bool random)
 
             if (random)
             {
-                var byteCount = Base64CharsToBytes(length);
-                var data = new byte[byteCount];
+                int byteCount = Base64CharsToBytes(length);
```
michaelcfanning, line 94: >int [](http://example.com/codeflow?start=16&length=3)

this conversion looks right to me, as retval type isn't explicit (though we're not surprised by it)

src/Microsoft.Security.Utilities.Cli/ExportDetectionsCommand.cs
```diff
@@ -24,7 +24,7 @@ internal int Run(ExportDetectionsOptions options)
                      "PreciselyClassifiedSecurityKeys.json",
                       WellKnownRegexPatterns.PreciselyClassifiedSecurityKeys);
 
-            foreach (var precision in new[] { DetectionMetadata.HighConfidence, DetectionMetadata.MediumConfidence, DetectionMetadata.LowConfidence })
+            foreach (DetectionMetadata precision in new[] { DetectionMetadata.HighConfidence, DetectionMetadata.MediumConfidence, DetectionMetadata.LowConfidence })
```
michaelcfanning, line 27: >DetectionMetadata [](http://example.com/codeflow?start=21&length=17)

i'm a very difficult person, but to me the type information in this data is clear because the enum value is encoded everywhere.

i'm pretty confident no set of style settings will satisfy my own list of do's and don'ts and automating this is a key value.

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: I believe it's because it has a separate knob for built-in types. And my choices are built-in = var or built-in = explicit and I don't know how to say built-in = defer to other knob. :(

michaelcfanning: ok, i suppose setting `csharp_style_var_for_built_in_types` results in wild, perhaps objectionable code churn? i just want any `new` utilization to reduce to a var decl. i can live with what we have though!

michaelcfanning (approved):
:shipit:

src/Microsoft.Security.Utilities.Benchmarks/SecretMaskerDetectionBenchmarks.cs
nguerrera: https://github.com/dotnet/roslyn/issues/26531 is related so we're not the only ones picking on this.

nguerrera: oh hmm, while the issue is still open the last comment suggests there's some magic we can use. Trying.

nguerrera: Nevermind, misread, that's a proposal, not implemented. I'll put a thumbs up.

PR #213, opened 2025-05-15
Add VS Code configuration

nguerrera:


.vscode/settings.json
```diff
@@ -0,0 +1,17 @@
+{
+  "dotnet.automaticallyCreateSolutionInWorkspace": false,
+  "dotnet.defaultSolution": "src/Microsoft.Security.Utilities.sln",
+  "editor.insertSpaces": true,
+  "editor.tabSize": 2,
```
michaelcfanning, line 5: >tabSize [](http://example.com/codeflow?start=10&length=7)

SEV6: not strictly required and I'm not sure this would be the default anyway. I do like tab size 2 for XML

michaelcfanning (approved):
:shipit:

PR #214, opened 2025-05-15
Fix switch case formatting

nguerrera:
* Unbraced `case` -> indent.
* Braced `case` -> do not indent.

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -232,6 +232,7 @@ csharp_style_throw_expression = true:suggestion
 csharp_indent_labels = no_change
 csharp_space_around_binary_operators = before_and_after
 csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = false
 csharp_indent_braces = false
 
 [*.vb]
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -231,7 +231,7 @@ csharp_style_expression_bodied_local_functions = false:silent
 csharp_style_throw_expression = true:suggestion
 csharp_indent_labels = no_change
 csharp_space_around_binary_operators = before_and_after
-csharp_indent_case_contents = false
+csharp_indent_case_contents = true
 csharp_indent_braces = false
 
 [*.vb]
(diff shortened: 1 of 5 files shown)


src/Microsoft.Security.Utilities.Core/Marvin.NetFramework.cs
```diff
@@ -112,32 +112,32 @@ public static long ComputeHash(byte[] data, ulong seed, int offset, int length)
                 switch (length)
                 {
                     case 0:
-                    p0 += 0x80u;
-                    break;
+                        p0 += 0x80u;
```
michaelcfanning, line 115: >+= [](http://example.com/codeflow?start=27&length=2)

What *should* have fired here is that we require case statements to be braced. I think that's why we don't require indentation.

This is what we want.

                    case 0:
                    {
                        p0 += 0x80u;
                        break;
                     }
 #Resolved

src/.editorconfig
```diff
@@ -231,7 +231,7 @@ csharp_style_expression_bodied_local_functions = false:silent
 csharp_style_throw_expression = true:suggestion
 csharp_indent_labels = no_change
 csharp_space_around_binary_operators = before_and_after
-csharp_indent_case_contents = false
+csharp_indent_case_contents = true
```
michaelcfanning, line 234: >true [](http://example.com/codeflow?start=30&length=4)

your change is ok is it doesn't indent case contents that are braced. can we apply braces to the modified code and check that? #Resolved

michaelcfanning:
formatting is a mysterious endeavor but fortunately not a topic that troubles me unduly when behavior is unexplained. as opposed to nearly everything else...

src/Microsoft.Security.Utilities.Core/Marvin.NetFramework.cs
nguerrera: Ah, I see. I can't find a way to enforce braces, but I did find a way to make it format sensibly (IMHO) with or without braces. 

See latest changes.

nguerrera:
It's explained now, thanks! I didn't change any code manually, but with my latest change, I now see improvements in the diff to existing code with and without braces, so I think we're good.

michaelcfanning (approved):
:shipit:

PR #216, opened 2025-06-05
Do not send duplicate detections to MaskSecrets callback

nguerrera:
While we still send detections that are nested in other detections, partially overlap other detections, or are adjacent to other detections, this change no longer sends more than one detection with the exact same start position and length. The main motivation here is for the use case where there are literal values added to the masker that also match regexes. The literal value is a known secret and may commonly also match a regex that finds secrets of its kind. The regex match is secondary in this case and not something that the caller will want to act upon.

This change also improves the reliability of preferring literal redaction token over regex redaction token by not having it rely on `"***"` (default literal) sorting before `"+++"` (default regex) and `"SECNNN/NNN:C3ID"` (high entropy rule regex with generateCorrelatingIds = true). The default `"***"` and `"+++"` tokens could be customized such that this would stop working. The sort order now explicitly prioritizes literal matches over regex matches, irrespective of redaction tokens. We also sort by ID before sorting by redaction token so that the detections sent to the callback do not change depending on generateCorrelatingIds = true or false.


src/Microsoft.Security.Utilities.Core/DetectionKind.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Detected by literal value.
+    /// </summary>
+    Literal = 0,
+
+    /// <summary>
+    /// Detected by regular expression.
+    /// </summary>
+    Regex = 1,
```
cfaucon: nit: based on the usage of this enum it seems Regex is the default.  Is there a reason why it is not in the 0 value here as well?

nguerrera: The reason is this value is used to sort literal matches before regex matches. The default argument was chosen more opportunistically because there are many places where we produce regex matches and only one where we produce literals.

I could reverse the order here and adjust the comparer accordingly. 

I'll do that and add some comments.

cfaucon: Or is this intentional I am assuming so that the compare favors literals first

cfaucon (approved):

src/Microsoft.Security.Utilities.Core/DetectionKind.cs
nguerrera: Right, it was, but I found a way to address both concerns and pushed a change that should help future us looking at this code. :)

cfaucon (approved):

PR #219, opened 2025-06-11
Re-normalize line endings and keep them normalized using .gitattributes

nguerrera:
I noticed in #218 that it looks like every line is changing in `SEC101_200_UnclassifiedLegacyCommonAnnotatedSecurityKey.cs`.

This has happened before and this solves the mystery and should prevent it from happening again.

Reviewing with ignore whitespace should only show the new .gitattributes change since all other changes are re-normalized line endings.


michaelcfanning (approved):
:shipit:

PR #220, opened 2025-06-11
Use actual namespace as RootNamespace

nguerrera:
This is breaking the build if you let VS regenerate the strongly-typed resource code in #218.

The namespace does not actually have Core in it.


michaelcfanning (approved):
:shipit:

PR #221, opened 2025-06-17
Test improvements

nguerrera:
Improve iteration time:
* Fix ValidateSecurityModelsMasking looping over regex engines when it is passed a regex engine.
* Limit testing with high-performance scanner disabled to the test that compares behavior of high-performance on and off.
* Stop using DetectSecrets on top of MaskSecrets in ValidateSecurityModelsMasking.
* Don't match more than once to get standalone secret.
* Full test run drops from ~35 seconds to ~9 seconds on my machine.

Improve coverage:
* Also test with false positive examples wherever true positive examples are used.
* Fix and use test expectations in SecretMasker_PlaceholderTestSecretsAreMasked.

Cleanup:
* Delete SecretMasker_PotentialSymmetricKeysAreClassified that had no implementation.


nguerrera:
Replaced by #222 from non-fork so I can stack a PR on top of it.

PR #223, opened 2025-06-18
Add DetectSecrets and MaskSecrets overloads that take `ReadOnlyMemory<char>`

### LINKED ISSUE #200: Add `SecretMasker.DetectSecrets(ReadOnlyMemory<char>)`
Partner is getting flagged for large allocation to call secret masker and requests an API to pass a span. (We have already planned to add this, but there was no existing issue tracking it.)

The biggest complexity here is the RegexEngine abstraction operates on strings today.

Also, there is no support in .NET Framework or .NET Standard for scanning spans with regexes.

nguerrera:
Fix #200

Internally, this is implemented using a `StringInput` struct that is backed by `ReadOnlyMemory<char>` on modern .NET and `string` on .NET Framework/Standard.

This is supported only on modern .NET 8+. Custom regex engines will throw if used with this API unless they are updated to implement a new Matches overload that takes `Span<char>`.

`RegexPattern.GetDetections` and `SecretLiteral` are now internal. These API needed breaking changes to implement this, and it will be better to keep them as implementation details moving forward.


src/Microsoft.Security.Utilities.Core/IRegexEngine.cs
```diff
@@ -10,5 +10,31 @@ namespace Microsoft.Security.Utilities
...
+    internal static class RegexEngineExtensions
+    {
+        public static IEnumerable<UniversalMatch> Matches(this IRegexEngine engine, StringInput input, string pattern, RegexOptions? options = null, TimeSpan timeout = default, string captureGroup = null)
+        {
+#if NET
+            if (input.TryGetString(out string s))
+            {
+                return engine.Matches(s, pattern, options, timeout, captureGroup);
+            }
+            else
```
michaelcfanning: >else [](http://example.com/codeflow?start=12&length=4)

>else [](http://example.com/codeflow?start=12&length=4)

SEV 8. I personally wouldn't create a braced code block here to emphasize that the previous branch never continues. #Closed

nguerrera: While SEV 8, I agree, so will do.

src/Microsoft.Security.Utilities.Core/StringInput.cs
```diff
@@ -0,0 +1,141 @@
...
+    // Implementation notes:
+    //
+    // * Making this a ref struct and using ReadOnlySpan in place of
+    //   ReadOnlyMemory would be better, but that requires significant
+    //   additional refactoring due to pervasive use of iterators (yield
+    //   return).
+    //
+    // * It is important that we do not retain the backing ReadOnlyMemory<char>
+    //   beyond the end of a DetectSecrets or MaskSecrets call. The caller owns
+    //   the memory. One of the disadvantage of using ReadOnlyMemory instead of
```
nguerrera:
```suggestion
    //   the memory. One of the disadvantages of using ReadOnlyMemory instead of 
``` #Closed

File: src/Tests.Microsoft.Security.Utilities.Core/RegexEngineTests.cs [hunk 3]
```diff
@@ -0,0 +1,103 @@
...
+            // Run the same test directly against the engine.
+            UniversalMatch match = engine.Matches(scanData, regex, captureGroup: "refine").FirstOrDefault();
+            match.Should().NotBeNull();
+            match.Success.Should().BeTrue();
+            match.Index.Should().Be(1); // 'x' at the start
+            match.Length.Should().Be(nameof(scanData).Length);
+            match.Value.Should().Be(nameof(scanData));
+        }
+
+
```
nguerrera, line 75: extra line of whitespace here.

src/Tests.Microsoft.Security.Utilities.Core/RegexEngineTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            RegexEngine_RefineGroup(RE2RegexEngine.Instance);
+        }
+
+        [TestMethod]
+        public void RegexEngine_CachedDotNet_RefineGroup()
+        {
+            RegexEngine_RefineGroup(RE2RegexEngine.Instance);
+        }
+
+        private static void RegexEngine_RefineGroup(IRegexEngine engine)
```
nguerrera, line 27: There should be a RefineGroupUsingSpan test #Closed

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
```diff
@@ -79,17 +79,23 @@ public List<HighPerformanceDetection> Scan(StringInput input)
             return [];
         }
 
+#if NET
+        HighPerformanceStringInput highPerformanceInput = input.Span;
```
nguerrera, line 85: It's tempting to use StringInput throughout in this file, but that leads to really chatty Memory -> Span conversion because we do things like look at a single character in the input. Probably worth a comment. 

If we could move everything to Span from Memory, this is another thing we wouldn't have to worry about.

src/Tests.Microsoft.Security.Utilities.Core/SecretMaskerTests.cs
```diff
@@ -477,13 +561,31 @@ public void SecretMasker_Encoder_UriDataEscape_LargeString()
...
+
+    [TestMethod]
+    public void SecretMasker_ProvidesVersion()
+    {
+        var expected = new Version(FileVersionInfo.GetVersionInfo(typeof(SecretMasker).Assembly.Location).FileVersion);
+        Version actual = SecretMasker.Version;
+        Assert.AreEqual(expected.Major, actual.Major);
+        Assert.AreEqual(expected.Minor, actual.Minor);
+        Assert.AreEqual(expected.Build, actual.Build);
+        Assert.AreEqual(-1, actual.Revision); // SecretMasker.Version does not include revision.
```
nguerrera, line 588: Unrelated and not very important, but I was looking critically at code coverage and this was all that remained to get SecretMasker to 100% so why not. :)

src/Microsoft.Security.Utilities.Core/PotentialSecurityKeys/SEC000_001_Unclassified64ByteBase64String.cs
```diff
@@ -32,19 +32,19 @@ public override IEnumerable<string> GenerateTruePositiveExamples()
...
+        foreach (Detection detection in base.GetDetections(input, generateCrossCompanyCorrelatingIds, defaultRedactionToken, regexEngine))
         {
             string match = input.Substring(detection.Start, detection.Length);
 
-            if (!Equals(azure64ByteIdentifiableKeys.GetDetections(match,
-                                                                         generateSha256Hashes,
-                                                                         defaultRedactionToken,
-                                                                         regexEngine).FirstOrDefault(), objB: default))
+            if (azure64ByteIdentifiableKeys.GetDetections(match,
+                                                          generateCrossCompanyCorrelatingIds: false,
```
nguerrera, line 45: unrelated, but renamed generateSha256Hashes to generateCrossCompanyCorrelatingIds to match current API. Also used Any() isntead of FirstOrDefault equals Default. Finally, we don't need to generate cross-company correlating IDs if all we care about is Any() so we don't past that through. #Closed

docs/ReleaseHistory.md
```diff
@@ -12,6 +12,10 @@
 - FPS => False positive reduction in static analysis.
 - FNS => False negative reduction in static analysis.
 
+# UNRELEASED
+- BRK: `SecretLiteral` and `RegexPattern.GetDetections` are no longer public.
```
michaelcfanning, line 16: OK, so people can't instantiate and search using a single pattern instance anymore. This allows us to obscure the new string input magic that we wouldn't want people to have to struggle to construct anyway (or at least it avoids requiring code to special-case their input or otherwise try to make the string vs. readonlyspan<char> distinction more seamless). As a result, users now are required to initialize a full masker instance with their pattern(s) or interest to get results. That all seems fine and there's a sensible workaround if we, in fact, break people by dropping this visibility. #Closed

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
```diff
@@ -44,32 +44,91 @@ public IEnumerable<UniversalMatch> Matches(string input, string pattern, RegexOp
             Regex regex = GetOrCreateRegex(pattern, options);
```
michaelcfanning, line 44: NIT on doc comment in this file, missed oppty to `cref` the reference to `System.Text.RegularExpressions.Regex` #Closed

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
```diff
@@ -44,32 +44,91 @@ public IEnumerable<UniversalMatch> Matches(string input, string pattern, RegexOp
...
             foreach (Match m in regex.Matches(input))
             {
-                yield return ToFlex(m, captureGroup);
+                yield return CreateUniversalMatch(captureGroup: captureGroup, match: m);
 
                 // Instance Regex.Matches has no overload; check timeout between matches
-                // (MatchesCollection *is* lazily computed).
+                // (MatchCollection *is* lazily computed).
```
nguerrera, line 50: I'm not loving this timeout. There's no way to pass it all the way to here when using SecretMasker. Also, it's only *between* regexes and I find silently dropping remaining matches vs. throw surprising. Also RE2 engine (in our tests) doesn't use it in the code path with refine group (the only one we use). I thought about ripping out timeout, but decided against unrelated breaking change on top of the related ones. :) #Closed

src/Microsoft.Security.Utilities.Core/SecretMasker.cs
```diff
@@ -274,6 +286,19 @@ public void AddLiteralEncoder(LiteralEncoder encoder)
...
+#if NET
+    public IEnumerable<Detection> DetectSecrets(ReadOnlyMemory<char> input)
+    {
+        return DetectSecrets(new StringInput(input));
+    }
+#endif
+
+    private IEnumerable<Detection> DetectSecrets(StringInput input)
+    {
+
```
nguerrera: extra whitespace here

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
```diff
@@ -79,17 +79,23 @@ public List<HighPerformanceDetection> Scan(StringInput input)
...
             if (pattern == null)
             {
                 break;
             }
 
-            if (Match(pattern, input, index, out HighPerformanceDetection detection))
+            if (Match(pattern, highPerformanceInput, index, out HighPerformanceDetection detection))
             {
                 detections.Add(detection);
             }
```
michaelcfanning, line 103: OK, so if we find an actual match on line 100, does that mean we can advance the index by the length of the detection (rather than the length of the pattern.Signature). According to some constraints that are pretty firm, we should never see an overlapping detection, right?

OTOH, it's not clear what benefit we receive, except maybe in our pathological 'nothing but secrets' test data design to maximize compute spent in matching.

It'd be nice if there were at least some discussion in code on this nuance. Or we could build in some kind of assert/checking for debug builds that verifies this assumption. Or we could in fact assume the absence of overlaps.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
```diff
@@ -156,7 +162,7 @@ private static bool Match(CompiledHighPerformancePattern pattern,
     /// The performance of this method is highly sensitive to the version of
     /// .NET that is used.
```
michaelcfanning, line 170: Related to previous comment, sorry I can't comment directly on the other code location. We can see in the Match implementation that we farm regex processing to an engine that professes to return multiple matches, if present, but we only return the first one. So again, some kind of code discussion on this may be helpful.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
```diff
@@ -215,7 +221,7 @@ private static bool Match(CompiledHighPerformancePattern pattern,
     /// packing 3 and 4 ASCII characters into integers and looking them up in
     /// a dictionary indexed by packed signatures.
     /// </summary>
-    private CompiledHighPerformancePattern? GetPatternForSignature(StringInput input, int index)
+    private CompiledHighPerformancePattern? GetPatternForSignature(HighPerformanceStringInput input, int index)
     {
         if ((input.Length - index) < 4)
         {
```
michaelcfanning, line 234: Comment below on requirement that signature below comprises ASCII, we could actually have a more stringent check here as things like ASCII null, bell or even the pipe character would be unexpected. Because in our scanner the signature itself is part of the sensitive data, we could constrain these characters to ASCII chars that can actually be encoded in a secret. Starting with alphanumeric, but including the base64 special chars (URL safe vs. non-URL safe) and expanding to the URL non-reserved (which adds period and tilde) #Closed

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: The reason here is simple, I'll add a comment. We don't want the regex running off and trying to match anywhere else in the input than exactly where the signature was found. It wouldn't even work since the scoped regexes also take advantage of how they are called and skip over the signature.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Yeah, we should look at this. I filed #224. I'm reluctant to change this pre-existing code in this unrelated change. It's complicated enough that it should be thought about, changed, and tested independently.

I'm even questioning the pattern.Signature.Length advance. It seems inconsistent. Either we support overlapping or we don't, I think. If we had signatures ABCD and BCDE then this would not work either, right?

In general, we are lacking on test coverage with more than one match in the input. While looking at this might be a great time to add that.

src/Microsoft.Security.Utilities.Core/StringInput.cs
```diff
@@ -0,0 +1,141 @@
...
+/// backed by <see cref="ReadOnlyMemory{Char}" on modern .NET and by <see
+/// cref="string"/> on legacy .NET Standard/Framework.
+/// </summary>
+internal readonly struct StringInput
+{
+    // Implementation notes:
+    //
+    // * Making this a ref struct and using ReadOnlySpan in place of
+    //   ReadOnlyMemory would be better, but that requires significant
+    //   additional refactoring due to pervasive use of iterators (yield
```
michaelcfanning, line 27: are we going to change this? i.e., reconsider our use of iterators? or, assuming a far-flung future where we run only on modern .NET, etc., etc., will we want to retain this pattern? thoughts #Closed

src/Microsoft.Security.Utilities.Core/StringInput.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            str = s;
+            return true;
+        }
+
+        str = null;
+        return false;
+    }
+#else
+    public string String { get; }
```
michaelcfanning, line 95: my care factor is low but this use of the non-aliased `String` type name makes reading this code trickier for me. `StringData` maybe? with a corresponding `SpanData` elsewhere. truly care factor is low. #Closed

src/Tests.Microsoft.Security.Utilities.Core/RegexEngineTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        [TestMethod]
+        public void RegexEngine_RE2_NoCaptureGroup()
+        {
+            RegexEngine_NoCaptureGroup(RE2RegexEngine.Instance);
+        }
+
+        [TestMethod]
+        public void RegexEngine_CachedDotNet_NoCaptureGroup()
+        {
+            RegexEngine_NoCaptureGroup(RE2RegexEngine.Instance);
```
michaelcfanning: Using wrong engine kind?

src/Tests.Microsoft.Security.Utilities.Core/RegexEngineTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        private static void RegexEngine_NoCaptureGroup(IRegexEngine engine)
+        {
+            string scanData = $"x{nameof(scanData)}x";
+            string regex = $"x(?P<refine>{nameof(scanData)})x";
```
michaelcfanning: Am I missing something? This test is called no capture group, but it does have a capture group.

src/Microsoft.Security.Utilities.Core/StringInput.cs
nguerrera: Yes, I'd like to change this and reconsider our use of iterators, but I wasn't sure it was the best use of time right now to dig in on that.

So I would:

1. Make StringInput ref struct and backed by Span on modern .NET (#225)
2. If and when we drop .NET standard/Framework (🙏), delete StringInput and use ReadOnlySpan<char> directly everywhere.

Keeping StringInput internal means that we can do both without breaking changes. It will do no harm to have Memory<char> overload that calls .Span and forwards to Span overload so there won't be a need for any breaking changes from here on.

src/Microsoft.Security.Utilities.Core/StringInput.cs
nguerrera: It does look a little weird here, but I like the way it reads for the caller and Memory has Memory.Span so it felt natural to continue that trend.

I'll keep this.

src/Tests.Microsoft.Security.Utilities.Core/RegexEngineTests.cs
nguerrera: Yikes.

src/Tests.Microsoft.Security.Utilities.Core/RegexEngineTests.cs
nguerrera: I'm testing the code path where you don't pass a capture group *argument* to engine.Matches. That was actually my main motivation for going in here. The masker systematically passes "refine" so there's no other way to cover that path than a test targeted at the engine. 

Yes, the name could be better. Suggestions? `NoCpatureGroupArgument`?

nguerrera: I can also change the name of the capture group in the expression to `unused`

michaelcfanning: Ah, of course. So `NoCaptureGroupRequested` or `NoRefinmentRequested` might make sense. I think `NoCaptureGroupRequested` with a capture group of `unused` would make sense.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Note that the ASCII requirement is handled here, not as an optimization to avoid consulting the dictionary, but because we cannot fit 4 non-ASCII chars into an Int32 key to the dictionary.

I am skeptical that adding more complex checks here would help and it might even hurt. I'm going to leave this alone, but if you think we should look deeper, feel free to file an issue and I can look at it.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Actually, it wouldn't run off because we currently enforce anchoring to the beginning but we may end up relaxing that.

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
nguerrera: fixed

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Comment added.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: This might have some risk of false negative. I will look at it with some urgency.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
michaelcfanning: I understand your view and am fine with this. I can't help think about the constraints on the input data, though. If a signature by our design is a segment of the secret, by definition, it can't include things like `\0`. If I felt strongly about it, I might suggest adding some debug only asserts that might pop on something unexpected. I don't feel strongly about it.

src/Microsoft.Security.Utilities.Core/StringInput.cs
michaelcfanning: good plan

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
michaelcfanning: Makes sense, and obviously helpful to document in source.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
michaelcfanning: For sure worth looking at. And yes, you're right, in theory, if advancing past the match is an issue so is advancing past the current signature.

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
nguerrera: Separate and I'm not sure it's worth an issue.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: I looked at this and I have some pending changes. I don't believe the changes are urgent after further review. Also, they don't relate to this PR in that it's all pre-existing so I'd prefer to keep them separate. I will send out the changes in a separate PR after this is approved and merged.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: Leaving this.

src/Tests.Microsoft.Security.Utilities.Core/RegexEngineTests.cs
nguerrera: Fixed, but GitHub being cute and not marking it outdated.

src/Microsoft.Security.Utilities.Core/HighPerformanceScanner.cs
nguerrera: #228 fixes this and is stacked on this one and waiting for it to merge to come out of draft.

michaelcfanning (approved):
:shipit:

src/Microsoft.Security.Utilities.Core/CachedDotNetRegex.cs
nguerrera: Discussion offline concluded we should file an issue. So I did: #229

PR #228, opened 2025-06-25
Improve handling of overlapping matches in high-performance scanner

### LINKED ISSUE #224: Review overlapping handling in HighPerformanceScanner
See https://github.com/microsoft/security-utilities/pull/223#discussion_r2155029155

nguerrera:
Fix #224 

1. Skip over successful matches and don't scan for overlapping matches. Improves performance scenario with dense secrets in the input like secret classification. Close to 20% improvement on 'NoPrefix' benchmark case.
2. Don't skip over signatures on successful signature match alone. It turns out this was OK with the current patterns that we have, but we were on shaky ground and it would have been easy to add a pattern where this could lead to false negatives.
3. Improve test coverage for these cases and also to enforce other invariants that are assumed by the high-performance scanner.


nguerrera:
Bump.

nguerrera:
I forgot about this *long ago* and I think it was meant to be taken out of draft when #223 merged. It is defensive against a future issue so was never urgent, but I think it would be good to take.

nguerrera:
Hello friends, I closed this only because nobody looked at it and it's time to sever anything that looks like it should rank on my TODO when querying github. I no longer work at Microsoft.

Issue #232, opened 2025-07-29
Rust crate microsoft_security_utilities_core should not depend on msvc_spectre_libs


whiskeyjay:
As a library microsoft_security_utilities_core should not assume it will always be targeting MSVC on Windows hence the need to link to Spectre mitigated MSVC lib for security. The msvc_spectre_libs import should be used in a Windows EXE/DLL project instead of a generic library project.

https://github.com/microsoft/security-utilities/blob/05f5fc7f9afa990161e6f5da38b1d4f27a1a4fef/src/security_utilities_rust/Cargo.toml#L21

I'm trying to use this library in a WASM project and the msvc_spectre_libs is failing the build because the WASM project build target isn't compatible with MSVC:

```
error: failed to run custom build command for `msvc_spectre_libs v0.1.3`                                                                

Caused by:
  process didn't exit successfully: `C:\Users\user\source\credactor_config_gen\target\release\build\msvc_spectre_libs-3ba3f9980c6a1776\build-script-build` (exit code: 101)
  --- stderr

  thread 'main' panicked at C:\Users\user\.cargo\registry\src\pkgs.dev.azure.com-3d7ccbe0d2b75539\msvc_spectre_libs-0.1.3\build.rs:22:14:
  unsupported arch: wasm32
```

nguerrera:
Fixed by #233

PR #233, opened 2025-07-29
Remove msvc_spectre_libs dependency to increase build target compatibility

whiskeyjay:
Proposing a fix for https://github.com/microsoft/security-utilities/issues/232

As a generic library `security_utilities_rust` should not reference the msvc_spectre_libs cargo - it is the final consuming Rust project, like a Windows EXE or DLL project's responsibility to decide whether or not to link against the Spectre mitigated MSVC libs.

The fix removes the msvc_spectre_libs reference.


michaelcfanning (approved):
:shipit:

PR #235, opened 2025-09-16
Fix build issues

nguerrera:
1. Using LangVersion=14 with .NET SDK 10.x while targeting .NET 10.0 hits a breaking change with Enumerable.Reverse vs Span.Reverse.

See https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/breaking-changes/compiler%20breaking%20changes%20-%20dotnet%2010#enumerablereverse

Fix: Lower LangVersion to 13.

2. The NuGet.config for the SecurityUtilitiesApiUtilizationExample was incompatible with a wrapping environment that disallows public NuGet.org feed.

Fix: Stop clearing feeds and stop referencing NuGet.org there.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup Label="Build">
     <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>
     <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>13</LangVersion>
     <OutputSubDir>$(Platform)_$(Configuration)</OutputSubDir>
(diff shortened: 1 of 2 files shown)


michaelcfanning:
> 		<add key="globalPackagesFolder" value="obj\packages" />

can we get rid of this?

---
Refers to: src/SecurityUtilitiesPackageReference/SecurityUtilitiesApiUtilizationExample/Nuget.config:14 in f8c5916. [](commit_id = f8c5916806cb4fc2142f3e228b9d874d90089dbc, deletion_comment = False)

michaelcfanning (approved):
:shipit:

nguerrera:
It's non-trivial to get rid of it vs. how we do the package-reference test. I will think about it, but for now, I am going with the minimal thing that will unblock builds.

michaelcfanning:
Of course, obviously you didn't make this change. And I marked your PR as approved. Next time I will be more explicit that I don't expect action, sorry about that.

nguerrera:
It's OK, I understood, hence why I merged. 😊

